PR #3, opened 2021-08-27
Add a unique_id category for pseudonymous identifiers

NevilleS:
This covers the (very common) case of assigning a `uuid` or similar to a user within your system. Making this an explicit category within `user_identifiable_data` parent should make it clear how to annotate this in your data!


cilliankieran (approved):
Perfect.

PR #4, opened 2021-09-10
Add the other 3 taxonomies

ThomasLaPiana:
* [x] add data subjects
* [x] add data uses
* [x] add data qualifiers


ThomasLaPiana:
@NevilleS copy pasta from the Fides repo

data_subjects.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: "Shareholder"
+
+  - fidesKey: "supplier_vendor"
+    name: "Supplier/Vendor"
+
+  - fidesKey: "trainee"
+    name: "Trainee"
+
+  - fidesKey: "visitor"
+    name: "Visitor
```
NevilleS: @ThomasLaPiana there's a bug here, malformed YAML missing a trailing quote. I've fixed in this repo but 👀 in case it surprises you elsewhere

ThomasLaPiana: thank you 👀

Issue #44, opened 2022-05-09
Enforce at least one "tag" set on each system


ThomasLaPiana:
### Is your feature request related to a specific problem?

Once #43 is implemented, we would like to enforce that each system has at least one "tag"

### Describe the solution you'd like

Add a pydantic validator to make sure that the "tag" isn't left empty for any systems

### Describe alternatives you've considered, if any

_A description of any alternative solutions or features you've considered._

### Additional context

_Add any other context or screenshots about the feature request here._

ThomasLaPiana:
@NevilleS we are able to surgically enforce tag existence requirements on systems only, if that's what we're looking for

NevilleS:
Personally, I'd like to make this something we elect to enforce only in the UI layer. If we were to make it a language requirement, I'd at least want a config variable to turn it on and off. For example, if we turned this on now, we'd break all existing fidesctl deployments 😄

ThomasLaPiana:
very correct! Would you like to close this for now? @NevilleS

NevilleS:
Yes, happy to close this one. If we do this, we'll focus on a layer above fideslang

PR #45, opened 2022-05-10
Add a `tags` field to all top-level fides resources

### LINKED ISSUE #43: Add a `tags` field to all top-level fides models

We want to let users "tag" their resources to eventually make filtering and reporting easier

We should add a `tags` field to the `FidesBase` model found [here](https://github.com/ethyca/fideslang/blob/main/src/fideslang/models.py#L38) as type `List[str]`

Adding this to the generic `meta` field instead

### Additional context

_Add any other context or screenshots about the feature request here._

ThomasLaPiana:
Closes #43

### Code Changes

* [x] add `tags` as a `Optional[List[str]]` field on `FidesModel` 
* [x] add tests

### Steps to Confirm

### Pre-Merge Checklist

* [x] Documentation Updated
* [x] Relevant Follow-Up Issues Created
* [x] Update `CHANGELOG.md`

### Description Of Changes

This PR adds the `tags` field to all top-level fides resources for eventual resource filtering/grouping

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,13 @@ The types of changes are:
 * `Fixed` for any bug fixes.
 * `Security` in case of vulnerabilities.
 
-## [Unreleased](https://github.com/ethyca/fideslang/compare/1.0.0...main)
+## [Unreleased](https://github.com/ethyca/fideslang/compare/0.9.0...main)
+
+### Added
+
+* There is now a `tags` field on the `FidesModel` model [#45](https://github.com/ethyca/fideslang/pull/45)
+
+## 0.9.0
 
 ### Added
 
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -45,6 +45,7 @@ class FidesModel(BaseModel):
         default="default_organization",
         description="Defines the Organization that this resource belongs to.",
     )
+    tags: List[str]
     name: Optional[str] = name_field
     description: Optional[str] = description_field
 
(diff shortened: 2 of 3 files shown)


ThomasLaPiana:
@NevilleS since we're going to require that every top-level has a tag, should we supply one as a default? Or are we specifically forcing people to tag their resources here?

Consequently, if we include a `default` tag, is that so different than not requiring it at all?

NevilleS:
I'm not convinced we should mandate a tag; that'll effectively prevent automated generation of valid resources, for example. If we programmatically add a `default` tag it pretty much defeats the purpose. Let's not require that yet

ThomasLaPiana:
@NevilleS from slack you specifically requested #44 , are we walking this back or saying that requirements should only exist for systems?

NevilleS:
Ah, from Slack I wanted to created that issue #44, but I think we can do this first (#43). I think that enforcing #44 needs some more discussion and refinement before committing to it

ThomasLaPiana:
@NevilleS sounds good, i'll make this optional then and we can discuss #44 in that issue

sanders41 (approved):
LGTM!

ThomasLaPiana:
@sanders41 thanks paul!
@NevilleS can you confirm this looks good to you?

NevilleS (approved):
LGTM. Not the most *expansive* test suite I've ever seen, but the implementation looks sound.

Would like to see a couple tests in `fidesctl` that exercise some cases like:
* adding a `tag` to an existing resource, running `apply`, and confirming that it's returned via the API
* removing a `tag` from a list of tags, running `apply`, and confirming that the tag is no longer returned
* etc.

ThomasLaPiana:
@NevilleS we can add those as part of the tests when we integrate this new version of fideslang

NevilleS:
Yes, of course and that would make sense. The language spec itself shouldn't have much logic.

PR #68, opened 2022-08-10
Add an is_default field to taxonomy fields

allisonking:
Part 1 of https://github.com/ethyca/fides/issues/854

### Code Changes

* [x] Adds a field `is_default` which defaults to false to the taxonomy fields (DataCategory, DataUse, DataSubject, DataQualifier)

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`

### Description Of Changes

Part 2 will take place in the fides repo and I believe will involve changes to the sql model for the same fields

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fideslang/compare/1.1.0...main)
 
+### Added
+
+* New field `is_default` added to DataCategory, DataUse, DataSubject, and DataQualifier [#68](https://github.com/ethyca/fideslang/pull/68)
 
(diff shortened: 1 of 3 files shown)


ThomasLaPiana:
@NevilleS are you ok with this change being here or do you want us to localize this only to fidesctl? Does this make sense to exist here in the broader sense?

NevilleS:
I think it's reasonable to include this in the core model, it certainly makes it easier to reason about the taxonomy as an end-user later on!

We'd need logic that prevents it ever being set to `True` at runtime though, right? By definition, the default taxonomy is static and defined here in the repo.

ThomasLaPiana (approved):

PR #85, opened 2022-09-16
Add `DataFlow`s, define `egress` and `ingress` on `System`s and `PrivacyDeclaration`s

### LINKED ISSUE #83: Add `ingress` and `egress` fields to the `system` resource for modeling data flows

We need a way to track directionality of data flowing between systems

Adding an `ingress` and an `egress` field to the `system` resources

More alternatives and discussion took place [here (internal only)](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2409398307/Data+Flow+Modeling)

### Additional context

_Current Definition_: `yaml system: - fides_key: demo_analytics_system name: Demo Analytics System description: A system used for analyzing customer behaviour. system_type: Service privacy_declarations: - name: Analyze customer behaviour for improvements. data_categories: - user.contact - user.device.cookie_id data_use: improve.system data_subjects: - customer data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified dataset_references: # This goes away - demo_users_dataset `_Proposed Definition_`yaml system: - fides_key: demo_analytics_system name: Demo Analytics System description: A system used for analyzing customer behaviour. system_type: Service privacy_declarations: - name: Analyze customer behaviour for improvements. data_categories: - user.contact - user.device.cookie_id data_use: improve.system data_subjects: - customer data_qualifier: aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified ingress: # This is added - fides_key: marketing_system type: system # An Enum of [system|dataset] action: process # An as-of-yet undefined Enum egress: # This is added - fides_key: customer_dataset type: dataset action: persist `

Note that the main changes here: 1. The removal of the `dataset_references` field 2. The addition of the `ingress` and `egress` fields

Design decisions to consider: 1. Ingress and Egress are defined on a per-declaration basis 2. This design does not let users specify which taxonomy types are involved in the egress/ingress, instead inheriting those values from the privacy declaration they are attached to 3. The enum values for `action` will need to be defined (process? persist? Read/write?) 4. No delineation between internal/external ingress/egress (external systems should be defined the same as internal systems, so this seems like an ok omission)

PSalant726:
Closes #83

### Code Changes

* [x] Define `egress` and `ingress` `FidesKey`s on `PrivacyDeclaration`s
* [x] Define the `DataFlow` resource model, and the companion `FlowableResources` `Enum`
* [x] Define `egress` and `ingress` `DataFlow`s on `System` resources
* [x] Validate that any `PrivacyDeclaration`s applied to a given `System` only reference `egress` and `ingress` `FidesKey`s that are also defined in that System's `egress` and `ingress` `DataFlow`s, respectively
* [x] Validate that any `DataFlow`s which set their `fides_key` and/or `type` field(s) to "user" also set their other field to "user"
* [x] Deprecate `PrivacyDeclaration.dataset_references` and `System.system_dependencies`
* [x] Fix some broken stuff

### Steps to Confirm

* [ ] Run the tests (they should pass)

### Pre-Merge Checklist

* [x] Documentation Updated
* [ ] ~Relevant Follow-Up Issues Created~
* [x] Update `CHANGELOG.md`

### DIFF
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 from enum import Enum
 from typing import Dict, List, Optional
+from warnings import warn
 
 from pydantic import AnyUrl, BaseModel, Field, HttpUrl, root_validator, validator
 
@@ -786,6 +787,21 @@ class System(FidesModel):
                     assert fides_key in [
                         data_flow.fides_key for data_flow in data_flows
                     ], f"PrivacyDeclaration '{value.name}' defines {direction} with '{fides_key}' and is applied to the System '{system}', which does not itself define {direction} with that resource."
+
+        return value
+
+    @validator("system_dependencies")
+    @classmethod
+    def deprecate_system_dependencies(cls, value: List[FidesKey]) -> List[FidesKey]:
+        """
+        Warn that the `system_dependencies` field is deprecated, if set.
+        """
+
+        if value is not None:
+            warn(
+                "The system_dependencies field is deprecated, and will be removed in a future version of fideslang. Use the 'egress' and 'ingress` fields instead.",
+                DeprecationWarning,
+            )
 
         return value
 
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -645,6 +645,12 @@ class PrivacyDeclaration(BaseModel):
     dataset_references: Optional[List[FidesKey]] = Field(
         description="Referenced Dataset fides keys used by the system.",
     )
+    egress: Optional[List[FidesKey]] = Field(
+        description="The resources to which data is sent."
+    )
+    ingress: Optional[List[FidesKey]] = Field(
+        description="The resources from which data is received."
+    )
 
 
 class SystemMetadata(BaseModel):
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -716,6 +716,21 @@ class DataFlow(BaseModel):
         description="An array of data categories describing the data in transit.",
     )
 
+    @root_validator()
+    @classmethod
+    def user_special_case(cls, values: Dict) -> Dict:
+        """
+        If either the `fides_key` or the `type` are set to "user",
+        then the other must also be set to "user".
+        """
+
+        if values["fides_key"] == "user" or values["type"] == "user":
+            assert (
+                values["fides_key"] == "user" and values["type"] == "user"
+            ), "The 'user' fides_key is required for, and requires, the type 'user'"
+
+        return values
+
 
 class System(FidesModel):
     """
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -761,6 +761,34 @@ class System(FidesModel):
 
     _check_valid_country_code: classmethod = country_code_validator
 
+    @validator("privacy_declarations", each_item=True)
+    @classmethod
+    def privacy_declarations_reference_data_flows(
+        cls,
+        value: PrivacyDeclaration,
+        values: Dict,
+    ) -> PrivacyDeclaration:
+        """
+        Any `PrivacyDeclaration`s which include `egress` and/or `ingress` fields must
+        only reference the `fides_key`s of defined `DataFlow`s in said field(s).
+        """
+
+        for direction in ["egress", "ingress"]:
+            fides_keys = getattr(value, direction, None)
+            if fides_keys is not None:
+                data_flows = values[direction]
+                system = values["fides_key"]
+                assert (
+                    data_flows is not None and len(data_flows) > 0
+                ), f"PrivacyDeclaration '{value.name}' defines {direction} with one or more resources and is applied to the System '{system}', which does not itself define any {direction}."
+
+                for fides_key in fides_keys:
+                    assert fides_key in [
+                        data_flow.fides_key for data_flow in data_flows
+                    ], f"PrivacyDeclaration '{value.name}' defines {direction} with '{fides_key}' and is applied to the System '{system}', which does not itself define {direction} with that resource."
+
+        return value
+
     class Config:
         "Class for the System config"
         use_enum_values = True
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -665,6 +665,36 @@ class SystemMetadata(BaseModel):
     )
 
 
+class FlowableResources(str, Enum):
+    """
+    The resource types with which DataFlows can be created.
+    """
+
+    DATASET = "dataset"
+    SYSTEM = "system"
+    USER = "user"
+
+
+class DataFlow(BaseModel):
+    """
+    The DataFlow resource model.
+
+    Describes a resource model with which a given System resource communicates.
+    """
+
+    fides_key: FidesKey = Field(
+        ...,
+        description="Identifies the System or Dataset resource with which the communication occurs. May also be 'user', to represent communication with the user(s) of a System.",
+    )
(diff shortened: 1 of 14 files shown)


src/fideslang/models.py
```diff
@@ -645,6 +645,12 @@ class PrivacyDeclaration(BaseModel):
     dataset_references: Optional[List[FidesKey]] = Field(
         description="Referenced Dataset fides keys used by the system.",
     )
+    egress: Optional[List[FidesKey]] = Field(
+        description="The resources to which data is sent."
```
NevilleS: I think this description can be a bit more helpful- for example:
```
The list of resources to which a data is sent. This is a list of `fides_key` entries, each of which must match an entry on the parent `egress` field on the `System` itself
```

or similar

src/fideslang/models.py
```diff
@@ -645,6 +645,12 @@ class PrivacyDeclaration(BaseModel):
...
+    )
+    ingress: Optional[List[FidesKey]] = Field(
+        description="The resources from which data is received."
```
NevilleS: Same as above

src/fideslang/models.py
```diff
@@ -665,6 +671,36 @@ class SystemMetadata(BaseModel):
...
+
+
+class DataFlow(BaseModel):
+    """
+    The DataFlow resource model.
+
+    Describes a resource model with which a given System resource communicates.
+    """
+
+    fides_key: FidesKey = Field(
```
NevilleS, line 707: I don't think we should allow a `fides_key: user` to be a special case here, so I'd argue this should be an optional field.

src/fideslang/models.py
```diff
@@ -665,6 +671,36 @@ class SystemMetadata(BaseModel):
...
+        ...,
+        description="Identifies the System or Dataset resource with which the communication occurs. May also be 'user', to represent communication with the user(s) of a System.",
```
NevilleS, line 709: I think the docs should say "if `type: user`, the `fides_key` must be null"

tests/fideslang/test_models.py
```diff
@@ -1,32 +1,134 @@
...
                 registry_id=1,
-                meta={"some": "meta stuff"},
-                fides_key="test_system",
+                system_dependencies=[],
                 system_type="SYSTEM",
-                name="Test System",
                 tags=["some", "tags"],
+            )
+
+    def test_system_no_ingress(self) -> None:
```
NevilleS, line 180: Not seeing any tests for the special case of `type: user` and some of the validations we should enforce there

NevilleS (changes requested):
Couple notes and nits

src/fideslang/models.py
PSalant726: We discussed this special case and decided it was reasonable?

Separately, if we allow `None` here, then any `PrivacyDeclaration`s that want to contextualize themselves as data coming to/from users would need to include `None` in their `ingress` and/or `egress` arrays. This gets messy, as `None` is not a valid `FidesKey` (nor should it be). Even if `None` _were_ a valid `FidesKey`, including `None` elements in arrays results in unnecessary allocations, which would just get garbage collected. So, those same `PrivacyDeclaration`s would define their `None` egress/ingress in the YAML files, but the resulting resource models would be created as empty arrays.

If you don't like `user` as the "special case `FidesKey`" then we can always use something else, but `None` does not work.

tests/fideslang/test_models.py
PSalant726: Was this comment assuming that we would update the "special case `FidesKey`" for users to `None`? If not, what tests/validations would you expect to see?

conceptualshark (approved):
👍 Thanks for the docs updates!

src/fideslang/models.py
ThomasLaPiana: Agreeing with @PSalant726 here, having `user` as a special `fides_key` allows us to treat `user` as a first-class object, otherwise we'll complicate a lot of the logic here

src/fideslang/models.py
```diff
@@ -719,6 +761,34 @@ class System(FidesModel):
 
     _check_valid_country_code: classmethod = country_code_validator
 
+    @validator("privacy_declarations", each_item=True)
```
ThomasLaPiana, line 795: we would also need a check here that any resources with the type of `user` also have `user` as the fides_key, since we're now supporting that as a special case here

tests/fideslang/test_models.py
```diff
@@ -1,32 +1,134 @@
...
+                        dataset_references=[],
+                        egress=["test_system_2"],
+                        ingress=["test_system_3"],
+                        name="declaration-name",
+                    )
+                ],
                 registry_id=1,
-                meta={"some": "meta stuff"},
-                fides_key="test_system",
+                system_dependencies=[],
```
ThomasLaPiana: `system_dependencies` is currently used for building the data map, but this language change will make that obsolete, we should communicate this change with @mfbrown and plan how to execute that change over there

I think we should also deprecate that field in this release

ThomasLaPiana: Adding a validator that gives a `DeprecationWarning` if that field is set by the user seems like a good option

tests/fideslang/test_models.py
ThomasLaPiana: `test_user_ingress_has_user_fides_key` or something similar I'd think, and we'd need a validator to go along with it

ThomasLaPiana (commented):
Left some comments, specifically we need to think about how to deal with `system_dependencies` and `dataset_references` either here or in a future issue as those are now redundant
Also I think `ingresses` and `egresses` should be used as the proper plural forms

ThomasLaPiana:
[Kubernetes agrees its ingresses](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class)

src/fideslang/models.py
NevilleS: 🤔 OK I think I follow. I feel like the special `fides_key` of `"user"` might surprise us later - for example, it's technically a valid `data_category` in the default taxonomy.

tests/fideslang/test_models.py
NevilleS: Yeah- whether or not the fides_key is optional, there's definitely special logic for `user` that deserves some test cases

PSalant726: The existing `System` validator - `privacy_declarations_reference_data_flows` - would already validate the "user" case. I've added a test to confirm. I also added a validator to `DataFlow`s to guarantee that if either the `fides_key` or `type` is set to "user", then so would be the other.

PSalant726:
Created #86 to track the removal of `PrivacyDeclaration.dataset_references` and `System.system_dependencies`. It should be addressed once a release is created with the changes on this PR.

src/fideslang/models.py
ThomasLaPiana: Due to the fact that fides_keys are only unique per-resource, everywhere we reference a fides_key we also describe what type of resource it is, so this collision shouldn't happen unless we drastically change fideslang in the future, and at that time we can tackle this I'd think

ThomasLaPiana (approved):
Good change, good tests 👍

NevilleS (approved):
LGTM. I'm a little suspicious about deprecating `dataset_references`, only because it's tightly woven into the policy evaluation on fides today, but I agree that it overlaps enough with the ingress/egress declaration that we can find a way to make that work.

PR #89, opened 2022-09-29
Assert `DataFlow.type` as a `str`

### LINKED ISSUE #88: `DataFlow`s include invalid fields when exported as YAML
### Bug Description

When a `System` resource that includes `egress` and/or `ingress` `DataFlow`s are exported to YAML using [`write_manifests()`](https://github.com/ethyca/fideslang/blob/f45fa281b1302566f0b46830d19cbf95a3d97915/src/fideslang/manifests.py#L9-L21), the `type` field renders incoherently. Ex:

```yml
type: !!python/object/apply:fideslang.models.FlowableResources
- system
```

### Expected behavior

The `type` field should be written as a literal string value. Ex:

```yml
type: system
```

### Additional context

Rather than assert that `type` parameters must be a member of the `FlowableResources` `Enum`, assert that `type` is a literal `str`, and add a validator that asserts it's inclusion in the `FlowableResources` `Enum`.

PSalant726:
Closes #88

### Code Changes

* [x] `DataFlow.type` --> `str`
* [x] Validate that `DataFlow.type` values are included in the `FlowableResources` `Enum`
* [x] Add a test for the new validator

### Steps to Confirm

* [ ] Run the tests (they should pass)
* [ ] Create a `System` resource with `DataFlow`s, and export it as YAML

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### Description Of Changes

Resolves an issue where `System` resources that include lists of `DataFlow`s were exported to invalid YAML.

### DIFF
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -708,15 +708,15 @@ class DataFlow(BaseModel):
         ...,
         description="Identifies the System or Dataset resource with which the communication occurs. May also be 'user', to represent communication with the user(s) of a System.",
     )
-    type: FlowableResources = Field(
+    type: str = Field(
         ...,
-        description="Specifies the resource model class for which the `fides_key` applies. May be any of 'dataset', 'system', or 'user'.",
+        description=f"Specifies the resource model class for which the `fides_key` applies. May be any of {', '.join([member.value for member in FlowableResources])}.",
     )
     data_categories: Optional[List[FidesKey]] = Field(
         description="An array of data categories describing the data in transit.",
     )
 
-    @root_validator()
+    @root_validator(skip_on_failure=True)
     @classmethod
     def user_special_case(cls, values: Dict) -> Dict:
         """
@@ -730,6 +730,18 @@ class DataFlow(BaseModel):
             ), "The 'user' fides_key is required for, and requires, the type 'user'"
 
         return values
+
+    @validator("type")
+    @classmethod
+    def verify_type_is_flowable(cls, value: str) -> str:
+        """
+        Assert that the value of the `type` field is a member
+        of the `FlowableResources` Enum.
+        """
+
+        flowables = [member.value for member in FlowableResources]
+        assert value in flowables, f"'type' must be one of {', '.join(flowables)}"
+        return value
 
 
 class System(FidesModel):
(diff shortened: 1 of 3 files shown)


src/fideslang/models.py
```diff
@@ -731,6 +731,18 @@ def user_special_case(cls, values: Dict) -> Dict:
...
 
+    @validator("type")
+    @classmethod
+    def verify_type_is_flowable(cls, value: str) -> str:
+        """
+        Assert that the value of the `type` field is a member
+        of the `FlowableResources` Enum.
+        """
+
+        flowables = [member.value for member in FlowableResources]
```
ThomasLaPiana, line 742:
```suggestion
        try: FlowableResource(value)
```

ThomasLaPiana: no need for an assert here, can natively check validity using the type

PSalant726: True, but then I would still need an `except` clause that checks and re-raises a `ValueError` to fail the validation. I think this is more concise and produces a more helpful error message.

tests/fideslang/test_models.py
```diff
@@ -21,6 +21,10 @@ def test_dataflow_user_type_no_user_fides_key(self) -> None:
         with raises(ValueError):
             assert DataFlow(fides_key="test_system_1", type="user")
 
+    def test_dataflow_invalid_type(self) -> None:
+        with raises(ValueError):
```
ThomasLaPiana, line 25: my gut stays this would be an assertion error? but i guess its still passing 🤔

ThomasLaPiana (approved):
We've seen this bug before, sorry I didn't catch it in the last PR!

tests/fideslang/test_models.py
PSalant726: Correct - `assert` raises an `AssertionError` on failure, but pydantic is cool with that. From [the docs](https://pydantic-docs.helpmanual.io/usage/validators/):
> * validators should either return the parsed value or raise a `ValueError`, `TypeError`, or `AssertionError` (`assert` statements may be used).

It's when the validator fails that a `ValueError` with the error message is raised.

PR #95, opened 2022-12-08
Add DSR Concepts to Fideslang [#94]

### LINKED ISSUE #94: Upstream DSR and connector related metadata from Fides
The fides project has extended the fideslang models to included extra metadata for processing DSRs. Most notably the `FidesopsMeta` model. All of the changes made to fideslang models in the [dataset.py](https://github.com/ethyca/fides/blob/ff4a4ca7b674e1dbca7c611094dffdb1c47c0f16/src/fides/api/ops/schemas/dataset.py) should be upstreamed into the fideslang project. This is to standardize on what DSR metadata should look like. 

Acceptance Criteria
- Upstreamed all of the DSR fields from fides into fideslang
- Update the Taxonomy model to not strip out extra fields that are included in JSON payloads
- Update `fidesops_meta` to `fides_meta`

pattisdr:
Closes https://github.com/ethyca/fideslang/issues/94

### Code Changes

* [x] Move `FidesDatasetReference`, `FidesMeta`, `FidesCollectionKey`, and `CollectionMeta`, and `EdgeDirection` concepts from Fides -> Fideslang.
* [x] Rename Dataset.fidesctl_meta to Dataset.fides_meta - merging the ops and ctl concepts.  Currently this just has the `resource_id` field and a new `after` field that allows you to specify the Datasets which should precede the current Dataset when running a DSR
* [x] Add `DatasetCollection.fides_meta` which currently only holds an `after` field, allowing you to specify which collections should run before the current collection in a DSR
* [x] Add `fides_meta` fields to DatasetField which is a `FidesMeta` resource.  This has several attributes, including which fields the current field references or is referenced by,  if the current field can be queried directly using identity data supplied for a DSR, whether the field is a primary key, a fides-supported datatype and other less common attributes.
    * [x] Add additional DatasetField.fields validation for fields of `object` type, related to what data a DSR may return
    * [x] Add additional fides_meta validation
* [x] Add a backwards compatible mixin that allows `fidesops_meta` to be specified on Datasets/Collections/Fields, and this is turned into a `fides_meta` attribute on instantiation
* [x] Have FidesValidationError be a subclass of ValueError instead of Exception so Pydantic can pick this up and raise ValidationErrors that are handled properly in FastAPI (note: this was one of the reasons why we opted to override FidesKey in Fidesops. See the "FidesOpsKey" concept in Fides.
* [x] Add "<", ">" to allowed keys in FidesKey (need approval that this is OK) - this is useful in the saas datasets where we want to specify fields in yaml that are eventually replaced with other text

### Steps to Confirm

I think the best way to validate is to pull this fides branch down and QA it, which is specifically using this fideslang branch https://github.com/ethyca/fides/pull/2009

* [ ] First check out main. In the Admin UI, under connections http://localhost:3000/datastore-connection, create a Postgres Connection and a Saas connector like sentry.  Verify in the db that their `datasetconfig.datasets` uses `fidesops_meta`.  Run a privacy request
* [ ] Still in main, also go to the Datasets tab and create a Dataset http://localhost:3000/dataset.  Verify existence in `ctl_datasets` table
* [ ] Now switch to https://github.com/ethyca/fides/pull/2009 branch.  View the previously created ops dataset Postgres configs and saas configs.  View the previously created ctl dataset
* [ ] Run another privacy request that should use the previously saved connection configs with fidesops_meta
* [ ] Delete those connectors and create new ones - Verify in the db that their `datasetconfig.datasets` uses `fides_meta`. now
* [ ] Run a privacy request, verify output is what is expected. 
* [ ]  go to the Datasets tab and create another dataset http://localhost:3000/dataset.  Verify existence in `ctl_datasets` table
* [ ] Verify ctl_datasets table has fides_meta column instead of fidesctl_meta.
* [ ] Spot check various cli commands

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`

### Description Of Changes

Fidesops expanded Fideslang Dataset/Collection/Field concepts in its own repo primarily to add concepts to carry out DSR's.  The most important of these are the `DatasetField.fidesops_meta` information which summarizes things like how collections are related to each other or if a collection is an entrypoint (what identity data we should start with to query the collection).

Move several DSR concepts from Fides back to Fideslang.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,7 @@ parentdir_prefix =
 [mypy] 
 show_error_codes = True
 check_untyped_defs = True
-disallow_any_explicit = False
+disallow_any_explicit = True
 disallow_untyped_defs = True
 files = src/
 exclude = migrations/
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,7 @@ parentdir_prefix =
 [mypy] 
 show_error_codes = True
 check_untyped_defs = True
-disallow_any_explicit = True
+disallow_any_explicit = False
 disallow_untyped_defs = True
 files = src/
 exclude = migrations/
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -335,25 +335,11 @@ class FidesopsMetaBackwardsCompat(BaseModel):
     """Mixin to convert fidesops_meta to fides_meta for backwards compatibility
     as we add DSR concepts to fideslang"""
 
-    @root_validator(allow_reuse=True, pre=True)
-    @classmethod
-    def fidesops_meta_conversion(cls, values: Dict) -> Dict:
-        """
-        If fidesops_meta specified, move this to fides_meta instead.
-        """
-        fidesops_meta = values.get("fidesops_meta")
-        fides_meta = values.get("fides_meta")
-
-        if fidesops_meta and fides_meta:
-            raise ValueError(
-                "You cannot specify both fidesops_meta and fides_meta. Please use fides_meta."
-            )
-
-        if fidesops_meta:
-            values["fides_meta"] = fidesops_meta
-            values["fidesops_meta"] = None
-
-        return values
+    def __init__(self, **data):
+        super().__init__(
+            fides_meta=data.pop("fidesops_meta", None) or data.pop("fides_meta", None),
+            **data,
+        )
 
 
 class DatasetField(DatasetFieldBase, FidesopsMetaBackwardsCompat):
@@ -364,9 +350,7 @@ class DatasetField(DatasetFieldBase, FidesopsMetaBackwardsCompat):
     """
 
     fides_meta: Optional[FidesMeta] = None
-    fidesops_meta: Optional[
-        FidesMeta
-    ] = None  # Will be deprecated eventually in favor of fides_meta
+
     fields: Optional[List[DatasetField]] = Field(
         description="An optional array of objects that describe hierarchical/nested fields (typically found in NoSQL databases).",
     )
@@ -473,9 +457,6 @@ class DatasetCollection(FidesopsMetaBackwardsCompat):
     )
 
     fides_meta: Optional[CollectionMeta] = None
-    fidesops_meta: Optional[
-        CollectionMeta
-    ] = None  # Will be deprecated eventually in favor of fides_meta
 
     _sort_fields: classmethod = validator("fields", allow_reuse=True)(
         sort_list_objects_by_name
@@ -539,9 +520,6 @@ class Dataset(FidesModel, FidesopsMetaBackwardsCompat):
     fides_meta: Optional[DatasetMetadata] = Field(
         description=DatasetMetadata.__doc__, default=None
     )
-    fidesops_meta: Optional[DatasetMetadata] = Field(
-        description=DatasetMetadata.__doc__, default=None
-    )  # Will be deprecated eventually in favor of fides_meta
     joint_controller: Optional[ContactDetails] = Field(
         description=ContactDetails.__doc__,
     )
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -6,7 +6,7 @@ Contains all of the Fides resources modeled as Pydantic models.
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional, Union
 from warnings import warn
 
 from pydantic import (
@@ -310,19 +310,30 @@ class FidesDatasetReference(BaseModel):
 
 
 class FidesMeta(BaseModel):
-    """Annotations used for query traversal"""
+    """Supplementary metadata used by the Fides application for additional features."""
 
-    references: Optional[List[FidesDatasetReference]] = None
-    identity: Optional[str]
-    primary_key: Optional[bool]
-    data_type: Optional[str]
-    """Optionally specify the data type. Fides will attempt to cast values to this type when querying."""
-    length: Optional[PositiveInt]
-    """Optionally specify the allowable field length. Fides will not generate values that exceed this size."""
-    return_all_elements: Optional[bool]
-    """Optionally specify to query for the entire array if the array is an entrypoint into the node. Default is False."""
-    read_only: Optional[bool]
-    """Optionally specify if a field is read-only, meaning it can't be updated or deleted."""
+    references: Optional[List[FidesDatasetReference]] = Field(
+        description="Fields that current field references or is referenced by. Used for drawing the edges of a DSR graph.",
+        default=None,
+    )
+    identity: Optional[str] = Field(
+        description="The type of the identity data that should be used to query this collection for a DSR."
+    )
+    primary_key: Optional[bool] = Field(
+        description="Whether the current field can be considered a primary key of the current collection"
+    )
+    data_type: Optional[str] = Field(
+        description="Optionally specify the data type. Fides will attempt to cast values to this type when querying."
+    )
+    length: Optional[PositiveInt] = Field(
+        description="Optionally specify the allowable field length. Fides will not generate values that exceed this size."
+    )
+    return_all_elements: Optional[bool] = Field(
+        description="Optionally specify to query for the entire array if the array is an entrypoint into the node. Default is False."
+    )
+    read_only: Optional[bool] = Field(
+        description="Optionally specify if a field is read-only, meaning it can't be updated or deleted."
+    )
 
     @validator("data_type")
     @classmethod
@@ -335,7 +346,7 @@ class FidesopsMetaBackwardsCompat(BaseModel):
     """Mixin to convert fidesops_meta to fides_meta for backwards compatibility
     as we add DSR concepts to fideslang"""
 
-    def __init__(self, **data: Any) -> None:
+    def __init__(self, **data: Union[Dataset, DatasetCollection, DatasetField]) -> None:
         """For Datasets, DatasetCollections, and DatasetFields, if old fidesops_meta field is specified,
         convert this to a fides_meta field instead."""
         fidesops_meta = data.pop("fidesops_meta", None)
@@ -377,7 +388,7 @@ class DatasetField(DatasetFieldBase, FidesopsMetaBackwardsCompat):
 
     @validator("fields")
     @classmethod
-    def validate_object_fields(
+    def validate_object_fields(  # type: ignore
         cls,
         fields: Optional[List["DatasetField"]],
         values: Dict[str, Any],
@@ -387,6 +398,7 @@ class DatasetField(DatasetFieldBase, FidesopsMetaBackwardsCompat):
         - Additionally object fields cannot have data_categories.
         """
         declared_data_type = None
+        field_name: str = values.get("name")  # type: ignore
 
         if values.get("fides_meta"):
             declared_data_type = values["fides_meta"].data_type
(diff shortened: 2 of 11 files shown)


src/fideslang/models.py
```diff
@@ -285,23 +298,150 @@ class MyDatasetField(DatasetFieldBase):
...
+    @classmethod
+    def validate_object_fields(
+        cls,
+        fields: Optional[List["DatasetField"]],
+        values: Dict[str, Any],
+    ) -> Optional[List["DatasetField"]]:
+        """Two validation checks for object fields:
+        - If there are sub-fields specified, type should be either empty or 'object'
+        - Additionally object fields cannot have data_categories.
+        """
```
pattisdr: This new validation on `fields` has the potential to conflict with already-defined Datasets on the -ctl side.

We've added a constraint from fidesops where we didn't want nested fields at different levels to have conflicting data categories.  This is because when we run a DSR we filter the end results on data category before uploading the results to the user.  We didn't want there to be ambiguity as to which data category should prevail, so we enforced that data categories should be pushed to the lowest level of nesting.

src/fideslang/models.py
```diff
@@ -285,23 +298,150 @@ class MyDatasetField(DatasetFieldBase):
...
+            return meta_values
+
+        is_array: bool = bool(
+            meta_values.data_type and meta_values.data_type.endswith("[]")
+        )
+        if not is_array and meta_values.return_all_elements is not None:
+            raise ValueError(
+                "The 'return_all_elements' attribute can only be specified on array fields."
+            )
+        return meta_values
```
pattisdr, line 387: This validation was ported over from fidesops: the `return_all_elements` attribute is for specific handling of array fields in DSR's. This will not conflict with existing ctl-generated datasets, as the `fides_meta` attribute did not exist.

https://github.com/ethyca/fides/blob/5a485387d8af247ec6479e4115088cbbb8394d77/docs/fides/docs/getting-started/datasets.md

> Optional. For array entrypoint fields, specify whether the query should return/mask all fields, or just matching fields. By default, we just return/mask matching fields. Setting return_all_elements=true will return/mask the entire array.

src/fideslang/validation.py
```diff
@@ -22,13 +21,15 @@ class FidesKey(ConstrainedStr):
     A FidesKey type that creates a custom constrained string.
     """
 
-    regex: Pattern[str] = re.compile(r"^[a-zA-Z0-9_.-]+$")
+    regex: Pattern[str] = re.compile(r"^[a-zA-Z0-9_.<>-]+$")
```
pattisdr, line 24: I need to get buy-in on this one, but a lot of saas templates have "<instance_fides_key>" in them which is replaced by a user-specified fides key when the template is created. 

In Fidesops, we had subclassed FidesKey for FidesOpsKey that specifically allows the "<instance_fides_key>"  string as an exception.  I'd like to make FidesKey more general.  In general, I think allowing "<" ">" characters in our datasets to indicate text that will be replaced by other text is a good pattern.

src/fideslang/validation.py
```diff
@@ -3,17 +3,16 @@
...
 from pydantic import ConstrainedStr
 
 from fideslang.default_fixtures import COUNTRY_CODES
 
-
 VALID_COUNTRY_CODES = [country["alpha3Code"] for country in COUNTRY_CODES]
 
 
-class FidesValidationError(Exception):
+class FidesValidationError(ValueError):
```
pattisdr, line 15: The primary reason we'd originally overridden FidesKey in Fidesops for a FidesOpsKey is that the FidesValidationError raises a 500 exception.  We wanted it to raise a ValueError that would be picked up by validator functions and raise a 422 http exception instead.

This one change helps a lot of FidesKey errors in unified fides be handled more gracefully.

src/fideslang/validation.py
```diff
@@ -93,3 +94,46 @@ def check_valid_country_code(country_code_list: List) -> List:
...
+
+
+def valid_data_type(data_type_str: Optional[str]) -> Optional[str]:
+    """If the data_type is provided ensure that it is a member of DataType."""
+
+    parsed_date_type, _ = parse_data_type_string(data_type_str)
+    if not is_valid_data_type(parsed_date_type):  # type: ignore
+        raise ValueError(f"The data type {data_type_str} is not supported.")
+
+    return data_type_str
```
pattisdr: We brought over this data type validation (which is for fides_meta attributes, something that -ctl didn't have, so there shouldn't be a conflict).  However, I didn't bring the data category validation over.  Fidesops was validating that the data category on a Dataset, Collection, or DatasetField matched one of the data categories in the static default taxonomy.  But in Unified Fides, ops-code now has access to data categories in the database which could include user-defined data categories.

I discussed a little with @SteveDMurphy but I think dictating data category validation to fides makes sense.

src/fideslang/models.py
```diff
@@ -285,23 +298,150 @@ class MyDatasetField(DatasetFieldBase):
...
+
+    def __init__(self, **data: Any) -> None:
+        """For Datasets, DatasetCollections, and DatasetFields, if old fidesops_meta field is specified,
+        convert this to a fides_meta field instead."""
+        fidesops_meta = data.pop("fidesops_meta", None)
+        fides_meta = data.pop("fides_meta", None)
+        super().__init__(
+            fides_meta=fides_meta or fidesops_meta,
+            **data,
+        )
```
pattisdr: Old datasets with fidesops_meta specified at the Dataset, Collection, or DatasetField level will still work here.  On instantiation, if fidesops_meta is specified, it will be converted to fides_meta.

src/fideslang/models.py
```diff
@@ -285,23 +298,150 @@ class MyDatasetField(DatasetFieldBase):
...
+    field: str
+    direction: Optional[EdgeDirection]
+
+
+class FidesMeta(BaseModel):
+    """Annotations used for query traversal"""
+
+    references: Optional[List[FidesDatasetReference]] = None
+    identity: Optional[str]
+    primary_key: Optional[bool]
```
pattisdr: These are some of the primary fields we're adding from fidesops that let the user specify how collections are related to each other which inform how a DSR graph is built.

src/fideslang/models.py
NevilleS: Thank you for doing this! So helpful

src/fideslang/models.py
```diff
@@ -285,23 +298,150 @@ class MyDatasetField(DatasetFieldBase):
...
+    def valid_data_type(cls, value: Optional[str]) -> Optional[str]:
+        """Validate that all annotated data types exist in the taxonomy"""
+        return valid_data_type(value)
+
+
+class FidesopsMetaBackwardsCompat(BaseModel):
+    """Mixin to convert fidesops_meta to fides_meta for backwards compatibility
+    as we add DSR concepts to fideslang"""
+
+    def __init__(self, **data: Any) -> None:
```
SteveDMurphy: Instead of using `Any` is it possible to use these explicitly as the docstring has?
```suggestion
    def __init__(self, **data: Union[Dataset, DatasetCollection, DatasetField]) -> None:
```

src/fideslang/models.py
SteveDMurphy: Thanks for the heads up here @pattisdr !! I did some testing and this ended up causing some issues with the current `fides_cache` Dataset that all made sense to me 👍🏽 

It was a little tough for me to see which ones were problematic, and can see if being tough on a larger dataset. Would it be possible to include some breadcrumbs to the offending fields to help with any user issues? Or would that be better done where implemented?

src/fideslang/validation.py
SteveDMurphy: I can't think of a better way to do that and I know we have continued to loosed the requirements here as needed... am thinking this makes sense too 👍🏽

SteveDMurphy (approved):
Added a couple of comments but installed and tested this in fides for validation purposes. The only thing that might be nice to do is see if there is a way to reduce or remove the use of `Any` if possible but not something that should hold this back

Also great job on the backwards compatibility addition!! 🔥 🔥

pattisdr:
thanks so much for your comments @SteveDMurphy 🏆

TheAndrewJackson (approved):
This is awesome! Thanks for taking over this work. Everything worked for me. I was unsure of exactly how the ops specific validations would split up between fideslang and fides. I like how everything worked out. I love the thorough tests for the `fidesops_meta` mixin 🚀

ThomasLaPiana:
these steps to confirm are 🔥 thank you for being so thorough @pattisdr

src/fideslang/models.py
```diff
@@ -285,23 +298,150 @@ class MyDatasetField(DatasetFieldBase):
...
+class FidesDatasetReference(BaseModel):
+    """Reference to a field from another Collection"""
+
+    dataset: FidesKey
+    field: str
+    direction: Optional[EdgeDirection]
+
+
+class FidesMeta(BaseModel):
+    """Annotations used for query traversal"""
```
ThomasLaPiana:
```suggestion
    """Supplementary metadata used by the Fides application for additional features."""
```

src/fideslang/models.py
```diff
@@ -285,23 +298,150 @@ class MyDatasetField(DatasetFieldBase):
...
+    references: Optional[List[FidesDatasetReference]] = None
+    identity: Optional[str]
+    primary_key: Optional[bool]
+    data_type: Optional[str]
+    """Optionally specify the data type. Fides will attempt to cast values to this type when querying."""
+    length: Optional[PositiveInt]
+    """Optionally specify the allowable field length. Fides will not generate values that exceed this size."""
+    return_all_elements: Optional[bool]
+    """Optionally specify to query for the entire array if the array is an entrypoint into the node. Default is False."""
+    read_only: Optional[bool]
```
ThomasLaPiana:
```suggestion
    read_only: Optional[bool] = Field(description="Optionally specify to query for the entire array if the array is an entrypoint into the node. Default is False.")
```

I _think_ this is the right syntax 

I believe these should all be included as descriptions instead of code comments so we can use them later for auto-generated docs as well as other potential things

setup.cfg
```diff
@@ -9,7 +9,7 @@ parentdir_prefix =
 [mypy] 
 show_error_codes = True
 check_untyped_defs = True
-disallow_any_explicit = True
+disallow_any_explicit = False
```
ThomasLaPiana: I don't think this is required anymore given @SteveDMurphy 's comment? I'm always very hesitant to use `Any` and had blocked it here for a reason, as it is the same as not having type annotations

If it is really necessary, could we use a local `type: ignore` instead of this root-level change?

src/fideslang/models.py
```diff
@@ -358,14 +500,16 @@ class DatasetMetadata(BaseModel):
     """
     The DatasetMetadata resource model.
 
-    Object used to hold application specific metadata for a dataset
+    Object used to hold application specific metadata for a dataset,
+    including annotations to help with query traversal.
```
ThomasLaPiana:
```suggestion
    including annotations to help with query traversal.
```

Nit: I'm not sure we need to specify the usage of query traversal here, as in the future we could use this field for any number of things, so specifying a single current use-case seems a bit odd

src/fideslang/validation.py
ThomasLaPiana: I was wondering why our error catching wasn't working correctly before, thank you for this fix!!

src/fideslang/validation.py
ThomasLaPiana: Makes sense to me!

src/fideslang/validation.py
```diff
@@ -93,3 +94,46 @@ def check_valid_country_code(country_code_list: List) -> List:
...
+    if not type_string:
+        return None, False
+    idx = type_string.find("[]")
+    if idx == -1:
+        return type_string, False
+    return type_string[:idx], True
+
+
+# Data types that Fides is currently configured to handle
+data_type_names: Set[str] = {
```
ThomasLaPiana:
```suggestion
DATA_TYPE_NAMES: Set[str] = {
```

Nit: since its a global, I think it should be upper-cased so I know immediately where to look for it

pattisdr:
Thanks so much for your comments @TheAndrewJackson and @ThomasLaPiana. I'll address today!

src/fideslang/models.py
pattisdr: Thanks for the callout, I've added the offending field name to the error message @SteveDMurphy -

TheAndrewJackson (approved):

pattisdr:
So I think we can go ahead and merge this!  For fides to be able to use this, there are some slight tweaks needed but we are handling as part of the unified fides resources. Fides is pinned to a specific fideslang version so I think this should be fine.

PR #101, opened 2023-04-17
set privacy declarations to use orm_mode

### LINKED ISSUE #100: Set pydantic `orm_mode` config to `True` for systems and privacy declarations

As part of https://github.com/ethyca/fides/issues/3036, we'll be making a proper db table to store the `privacy_declarations` of `System`s. Since we've already got pydantic models for `PrivacyDeclaration`s, by updating the `System`, `PrivacyDeclaration` and `DataUse` pydantic models (defined here in `fideslang`) to have `orm_mode` set to `True`, this would minimize the amount of code changes in `fides` that we'd need to implement this update; instead we can rely mainly on pydantic's orm model mapping to treat the new `PrivacyDeclaration` DB (ORM) records similarly to how they were being handled before as JSON blobs that were validated against the pydantic models

- Add a
```
class Config:
        orm_mode = True
```
to the `System`, `PrivacyDeclaration` and `DataUse` pydantic models defined in `fideslang`.

    - The reason we'll be updating the `DataUse` model is because we'll add a 'hard' FK reference from `PrivacyDeclaration` to `DataUse` records, rather than relying on a 'soft' string-matched reference that existed previously

Instead of making this update, we could update more of our code in `fides` to translate DB/ORM instances of `PrivacyDeclaration`s into JSON blobs as they'd been previously, so that they can be validated properly by the pydantic model. IMO, going that route would introduce more customized logic that would be fragile to future changes.

The proposed updates here would be a more "out of the box" solution that should be more maintainable and less error prone. Probably significantly less code too!

Another alternative: could/should we create a base pydantic class for our fideslang pydantic models that defaults to `orm_mode = True`? we may want to update more of our pydantic models in a similar fashion moving forward, so this could save us some updates. I'd need to do a bit more investigation to look at potential downsides/regressions from this broader change...

### Additional context

We may look to make similar changes moving forward, since we may want to store more of our `fideslang` pydantic models in their own DB tables. How can we make this as future proof as possible while also not blowing things up? See base class suggestion above.

adamsachs:
Closes #100 

### Code Changes

* set `orm_mode = True` on the `PrivacyDeclaration` model

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`

### Description Of Changes

Needed for https://github.com/ethyca/fides/issues/3036.

I _may_ need some additional changes to `fideslang` as part of https://github.com/ethyca/fides/issues/3036, though I think this is all that's needed currently. ~@ThomasLaPiana or anyone else - do we have any quick way to develop `fides` against new `fideslang` code without actually pushing through a release?~ update - i was able to develop/test against these changes by locally updating the `requirements.txt` in `fides` to point to this specific git commit. things generally looked good, i don't foresee any additional changes needed on the `fideslang` side at the moment - but there's still a _slight_ chance that could change going forward!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ The types of changes are:
 ## [Unreleased](https://github.com/ethyca/fideslang/compare/1.3.2...main)
 
 ### Changed
+* Make `PrivacyDeclaration` use pydantic `orm_mode` [#101](https://github.com/ethyca/fideslang/pull/101)
+
+## [1.3.3](https://github.com/ethyca/fideslang/compare/1.3.2...1.3.3)
+
+### Changed
 * Make `PrivacyDeclation.name` optional [#97](https://github.com/ethyca/fideslang/pull/97)
 
 ## [1.3.2](https://github.com/ethyca/fideslang/compare/1.3.1...1.3.2)
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -824,6 +824,11 @@ class PrivacyDeclaration(BaseModel):
 
         return value
 
+    class Config:
+        """Config for the Privacy Declaration"""
+
+        orm_mode = True
+
 
 class SystemMetadata(BaseModel):
     """


ThomasLaPiana (approved):

ThomasLaPiana:
@adamsachs do you want me to merge this and cut a release? Looks good to me :)

adamsachs:
@ThomasLaPiana yes please! thanks 👍

adamsachs:
@ThomasLaPiana  i'll go ahead and merge now :) if the release is as simple as pushing a tag, i'm happy to do that. if it's a bit more involved, would you mind doing it? or i'm happy to follow a step list if there is one!

adamsachs:
@ThomasLaPiana how does this look? https://github.com/ethyca/fideslang/releases/edit/untagged-6f1b297db4bc256ef0d6

PR #103, opened 2023-04-25
add `id` field to `PrivacyDeclaration` model

### LINKED ISSUE #102: include `id` in `PrivacyDeclaration` model

Now that `PrivacyDeclaration`s have their own DB table/records, we should allow their `id` to be exposed on API responses so that it can be referenced by other operations (e.g. custom field references)

- include an `Optional[str]` field of `id` in the `PrivacyDeclaration` model

Instead of updating `fideslang` here, we could instead extend the `PrivacyDeclaration` model class in `fides` to include the additional field, and it would be used specifically in response payloads. This may be a bit more targeted, but it's a lot more code changes and i think it will be hard to follow, so i prefer this option here.

### Additional context

I _believe_ this will be needed to allow the `fidesplus` FE to be able to add custom fields that reference particular `PrivacyDeclaration`s. without this `id` being returned in our API responses, there's no way that the FE will be able to know the `id` of particular `PrivacyDeclaration`.

adamsachs:
Closes #102 

### Code Changes

* add `id` field to `PrivacyDeclaration` model

### Steps to Confirm

* we'll need to integrate this into `fides` to be able to confirm it adds the API response data we are looking for
    * i confirmed this works as expected by pinning `fides` to this `fideslang` commit in the `requirements.txt`. the API response to getting a system with privacy declarations looked like this (note the `id` fields):
 ```
{
  "fides_key": "test_system_1",
  "organization_fides_key": "default_organization",
  "tags": [],
  "name": "test system 1",
  "description": "",
  "registry_id": null,
  "meta": null,
  "fidesctl_meta": null,
  "system_type": "",
  "data_responsibility_title": "Controller",
  "egress": null,
  "ingress": null,
  "privacy_declarations": [
    {
      "id": "pri_eb5ad4c6-164a-47c1-a803-a84078b24dde",
      "name": "test",
      "data_categories": [
        "user"
      ],
      "data_use": "advertising",
      "data_qualifier": "aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified",
      "data_subjects": [
        "commuter"
      ],
      "dataset_references": [],
      "egress": null,
      "ingress": null
    },
    {
      "id": "pri_ce89b310-3015-4373-bc94-2180b63d270e",
      "name": "test 2",
      "data_categories": [
        "user.browsing_history"
      ],
      "data_use": "advertising",
      "data_qualifier": "aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified",
      "data_subjects": [
        "next_of_kin"
      ],
      "dataset_references": [],
      "egress": null,
      "ingress": null
    }
  ],
  "system_dependencies": [],
  "joint_controller": null,
  "third_country_transfers": [],
  "administrating_department": "Not defined",
  "data_protection_impact_assessment": {
    "is_required": false,
    "progress": null,
    "link": null
  }
}
```

### Pre-Merge Checklist

* [ ] Documentation Updated
* [ ] Update `CHANGELOG.md`

### Description Of Changes

See issue for more context around the change. Essentially, I think we need to return the `PrivacyDeclaration.id` property to be able to link `CustomField`s to `PrivacyDeclaration`s

### DIFF
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -783,6 +783,9 @@ class PrivacyDeclaration(BaseModel):
     to the privacy data types.
     """
 
+    id: Optional[str] = Field(
+        description="The database-assigned ID of the privacy declaration on the system. This is meant to be a read-only field."
+    )
     name: Optional[str] = Field(
         description="The name of the privacy declaration on the system.",
     )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ The types of changes are:
 ## [Unreleased](https://github.com/ethyca/fideslang/compare/1.3.2...main)
 
 ### Changed
+* Include an optional `id` field on `PrivacyDeclaration` model [#103](https://github.com/ethyca/fideslang/pull/103)
+
+## [1.3.4](https://github.com/ethyca/fideslang/compare/1.3.3...1.3.4)
+
+### Changed
 * Make `PrivacyDeclaration` use pydantic `orm_mode` [#101](https://github.com/ethyca/fideslang/pull/101)
 
 ## [1.3.3](https://github.com/ethyca/fideslang/compare/1.3.2...1.3.3)
--- a/tests/fideslang/test_models.py
+++ b/tests/fideslang/test_models.py
@@ -1,6 +1,6 @@
 from pytest import deprecated_call, mark, raises
 
-from fideslang import DataFlow, PrivacyDeclaration, System, Dataset
+from fideslang import DataFlow, Dataset, PrivacyDeclaration, System
 from fideslang.models import DatasetCollection, DatasetField
 
 pytestmark = mark.unit
@@ -30,6 +30,18 @@ class TestDataFlow:
 class TestPrivacyDeclaration:
     def test_privacydeclaration_valid(self) -> None:
         assert PrivacyDeclaration(
+            data_categories=[],
+            data_qualifier="aggregated_data",
+            data_subjects=[],
+            data_use="provide",
+            egress=[],
+            ingress=[],
+            name="declaration-name",
+        )
+
+    def test_privacydeclaration_valid_with_id(self) -> None:
+        assert PrivacyDeclaration(
+            id="pri_12345",
             data_categories=[],
             data_qualifier="aggregated_data",
             data_subjects=[],

### CODE REVIEW

File: src/fideslang/models.py [hunk 1]
```diff
@@ -783,6 +783,9 @@ class PrivacyDeclaration(BaseModel):
     to the privacy data types.
     """
 
+    id: Optional[str] = Field(
+        description="The database-assigned ID of the privacy declaration on the system. This is meant to be a read-only field."
+    )
```
pattisdr, line 788: Seems straightforward to me and we do have some test coverage here asserting that no id doesn't break it.  https://github.com/ethyca/fideslang/blob/main/tests/fideslang/test_models.py#L32.  We could add a test case for where a privacy declaration does have an id too

adamsachs: oh good call! thanks for pointing that out. i'll add in a test

pattisdr (approved):
Looks good to me, especially making this optional makes it easy to integrate into fides.

src/fideslang/models.py
NevilleS: How do you plan to prevent users of the API/YAML/etc editing that value? Is there a nice way to have Pydantic guard against that kind of footgun, if only to just defend against accidental errors?

pattisdr: Should we just add this on the `Fides` side then?  we often have multiple schemas, one for the "creation" schema and another that extends this, adding just the id for the response.

adamsachs: yeah, i thought about this.

i'm not really sure the best way to do this at the pydantic level, beyond splitting out a different model for request/response, which is what we commonly do. but that would be a pretty disruptive code change, since we make a lot of assumptions in our `crud` API code that associates DB models with a single pydantic `fideslang` model.

i could add in some extra checks in the system API code in `fides`, that feels like the least intrusive thing to do at this point. those endpoints are all special-cased right now anyway, so i don't think it'd be too hard to add in a handler that rejects any write requests that specify an `id`.

NevilleS: All good from me! Just keeping an eye on how that little inconsistency in fideslang might start to confuse the language...

adamsachs:
closing in favor of https://github.com/ethyca/fides/pull/3157 which implements a similar change but scoped within `fides`, and more targeted to only response models.

PR #109, opened 2023-05-23
import `Literal` from `typing_extensions` to work around pydantic bug

### LINKED ISSUE #108: fideslang: incompatibility with python < `3.10` caused by pydantic bug
### Bug Description

Corresponding ticket to https://github.com/ethyca/fides/issues/3358 which fixes the problem in `fides`. But we should fix it also here in `fideslang` to keep it compatible in its own right with python < `3.10`.

#### Steps to Reproduce

- try "using" `fideslang` in a python env < `3.10`

### Expected behavior

- projects/apps that run on python < `3.10` should be able to easily leverage `fideslang`

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Environment

* Version:
* OS:
* Python Version:
* Docker Version:

### Additional context

See https://github.com/ethyca/fides/issues/3358 for more context

adamsachs:
Closes #108.

Caused by https://github.com/pydantic/pydantic/issues/5821.

### Code Changes

* import `Literal` from `typing_extensions` to work around the pydantic bug introduced with typing_extensions version `4.6.0`

### Steps to Confirm

* CI passes

### Pre-Merge Checklist

* [ ] Documentation Updated
* [ ] Update `CHANGELOG.md`

### Description Of Changes

I decided on this workaround here in `fideslang` (see https://github.com/pydantic/pydantic/issues/5821#issuecomment-1559196859 for a list of potential workaround) for a couple of reasons:
- we only have a single use of `Literal`
- in `fideslang`, pinning the `typing_extensions` package to `4.5.0` (which is the workaround i'd decided on in `fides` https://github.com/ethyca/fides/pull/3357) could come with more, and less visible, consequences for "downstream" users of the package, if e.g. other applications need to have `typing_extensions` on a _higher_ version.
    - so instead of potentially getting in their way, a simple code patch here means we shouldn't have any impact on others using this package

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,7 +3,7 @@ ipython
 mypy==0.910
 packaging==20.9
 pre-commit==2.9.3
-pylint==2.6.0
+pylint==2.10.0
 pytest==6.2.2
 pytest-cov==2.11.1
 requests-mock==1.8.0
--- a/src/fideslang/manifests.py
+++ b/src/fideslang/manifests.py
@@ -17,7 +17,7 @@ def write_manifest(
     else:
         manifest = {resource_type: manifest}
 
-    with open(file_name, "w") as manifest_file:
+    with open(file_name, "w", encoding="utf-8") as manifest_file:
         yaml.dump(manifest, manifest_file, sort_keys=False, indent=2)
 
 
@@ -25,7 +25,7 @@ def load_yaml_into_dict(file_path: str) -> Dict:
     """
     This loads yaml files into a dictionary to be used in API calls.
     """
-    with open(file_path, "r") as yaml_file:
+    with open(file_path, "r", encoding="utf-8") as yaml_file:
         loaded = yaml.safe_load(yaml_file)
         if isinstance(loaded, dict):
             return loaded
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -6,7 +6,7 @@ Contains all of the Fides resources modeled as Pydantic models.
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 from warnings import warn
 
 from pydantic import (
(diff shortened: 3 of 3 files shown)


src/fideslang/models.py
```diff
@@ -20,6 +20,9 @@
     validator,
 )
 
+# import `Literal` from typing_extensions to work around https://github.com/pydantic/pydantic/issues/5821
+from typing_extensions import Literal
```
NevilleS, line 24: nit: If we think we'd prefer to revert this later, adding a follow-up issue, self-assigning it, and throwing that into a sprint backlog a month or two from now is a power move 👍

NevilleS (approved):
Thanks!

dev-requirements.txt
```diff
@@ -3,7 +3,7 @@ ipython
 mypy==0.910
 packaging==20.9
 pre-commit==2.9.3
-pylint==2.6.0
+pylint==2.10.0
```
adamsachs, line 6: we're on a very outdated version of `pylint` and it was causing errors with the new `from typing_extensions import Literal`, so i figured i'd update.

updating to `2.15.4` as we have in `fides` caused a host of other pylint warnings to crop up that we'd need to address, and i figured that'd be out of scope for this PR, so i found a "middle ground" 🤷

src/fideslang/manifests.py
```diff
@@ -17,15 +17,15 @@ def write_manifest(
     else:
         manifest = {resource_type: manifest}
 
-    with open(file_name, "w") as manifest_file:
+    with open(file_name, "w", encoding="utf-8") as manifest_file:
```
adamsachs, line 20: this and L28 change were needed because of the `pylint` update

src/fideslang/models.py
adamsachs: yup, i was thinking about the follow up for sure.

why do you think we'll be able to revert in a month or two though? potential fix coming from pydantic?

adamsachs: made https://github.com/ethyca/fideslang/issues/110

NevilleS (approved):
Thanks!

PR #113, opened 2023-06-02
make `meta` attribute on `System` and `Dataset` model more permissive

### LINKED ISSUE #111: Consider updating type annotation of `meta` attribute to be more permissive

For https://github.com/ethyca/fides/issues/3315, we've discussed leveraging the `meta` attribute on the `System` model/class to pass in some arbitrary metadata. For that use case, it could be useful to specify nested JSON objects (ideally of arbitrary depth), rather than the "flat" (i.e. depth of 1) objects/dictionaries that are required currently based on the type annotation of `Optional[Dict[str, str]]`. 

@TheAndrewJackson also noted (in https://github.com/ethyca/fides/issues/3315) that this was a requirement he needed for the webscanner demo.

Could we allow for nested JSON objects of arbitrary depth for the `meta` attributes (of `System` and probably of `Dataset` too, for consistency)? 

e.g.
```
{
  "saas_config": {
    "type":"stripe",
    "icon":"***",
  }
}
```
or  
```
{
  "web_scan": {
    "domains": [
      {
        "url": "***",
        "cookies": [
          "***",
          "***",
          "***",
        ]
      }
    ]
  }
}
```

I think this may be as simple as updating the type annotations to `Optional[Dict[str, Any]]`?

There may be some unforeseen implications of making this change -- is it possible that any dependent code (either in our, i.e. Ethcya's, code base or anyone else who uses `fideslang`) is depending on the stricter typing that's in place, and if so, would this change be disruptive?

Additionally, it would be good to understand the original motivation behind the stricter typing, if that was a deliberate decision.

If we judge that updating the type annotation on the `meta` property to be more permissive is **not** a good path forward, then we _could_ create another property that allows for nested dictionaries (of arbitrary length) for arbitrary metadata. This feels pretty messy, though - especially considering we already have a `fidesctl_meta` attribute (that holds more strictly-typed system metadata)...

### Additional context

This is motivated by the work to more explicitly tie `System`s to saas connector types. See  https://github.com/ethyca/fides/issues/3315 for some more detail on that use case.

cc @TheAndrewJackson @galvana

adamsachs:
Closes #111 

### Code Changes

* [x] make `meta` attribute on `System` and `Dataset` models more permissive to allow for nested dictionaries/objects

### Steps to Confirm

* [x] preliminary test of fides integration [here](https://github.com/ethyca/fides/pull/3463) - looks good

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`

### Description Of Changes

see #111 for more context!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,16 +14,27 @@ The types of changes are:
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
-## [Unreleased](https://github.com/ethyca/fideslang/compare/1.3.2...main)
+## [Unreleased](https://github.com/ethyca/fideslang/compare/1.3.4...main)
+
+### Changed
+
+- Make `meta` property of `System` and `Dataset` models more permissive [#113](https://github.com/ethyca/fideslang/pull/113)
+
+## [1.4.0](https://github.com/ethyca/fideslang/compare/1.3.4...1.4.0)
+
+### Changed
+
+- Updated the default data uses [#107](https://github.com/ethyca/fideslang/pull/107)
+
+### Removed
+
+- The `system_dependencies` field of `System` resources [#105](https://github.com/ethyca/fideslang/pull/105)
+
+## [1.3.4](https://github.com/ethyca/fideslang/compare/1.3.3...1.3.4)
 
 ### Changed
 
 - Make `PrivacyDeclaration` use pydantic `orm_mode` [#101](https://github.com/ethyca/fideslang/pull/101)
-- Updated the default data uses [#107](https://github.com/ethyca/fideslang/pull/107)
-
-### Remove
-
-- The `system_dependencies` field of `System` resources [#105](https://github.com/ethyca/fideslang/pull/105)
 
 ## [1.3.3](https://github.com/ethyca/fideslang/compare/1.3.2...1.3.3)
 
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -33,7 +33,7 @@ from fideslang.validation import (
     valid_data_type,
 )
 
-# Reusable components
+# Reusable Validators
 country_code_validator = validator("third_country_transfers", allow_reuse=True)(
     check_valid_country_code
 )
@@ -45,6 +45,7 @@ no_self_reference_validator = validator("parent_key", allow_reuse=True)(
     no_self_reference
 )
 
+# Reusable Fields
 name_field = Field(description="Human-Readable name for this resource.")
 description_field = Field(
     description="A detailed description of what this resource is."
@@ -53,8 +54,8 @@ is_default_field = Field(
     default=False,
     description="Denotes whether the resource is part of the default taxonomy or not.",
 )
-
 meta_field = Field(
+    default_factory=dict,
     description="An optional property to store any extra information for a resource. Data can be structured in any way: simple set of `key: value` pairs or deeply nested objects.",
 )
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
adamsachs, line 1: changelog needed some general hygiene, figured i'd just do it here

adamsachs:
@ThomasLaPiana could i get a quick look at this? should be pretty straightforward, i hope 😄 

i'd like to cut a point release with this change after it's merged, if that's alright, as i'm hoping to get it integrated over to `fides` (https://github.com/ethyca/fides/pull/3463) before the next fides release later this week.

src/fideslang/models.py
```diff
@@ -526,10 +526,10 @@ class DatasetMetadata(BaseModel):
     after: Optional[List[FidesKey]]
 
 
-class Dataset(FidesModel, FidesopsMetaBackwardsCompat):
+class Dataset(FidesModel, FidesopsMetaBackwardsCompat):  # type: ignore[misc]
     """The Dataset resource model."""
 
-    meta: Optional[Dict[str, str]] = Field(
+    meta: Optional[Dict[str, Any]] = Field(  # type: ignore[misc]
```
ThomasLaPiana: sorry to nit here, but I feel like this should be a reusable field since we're using it in at least two places and can move it to others later if we want.

See the top of the file for examples of reusable fields, you can add it there and update the description and this should be good to go!

src/fideslang/models.py
```diff
@@ -526,10 +526,10 @@ class DatasetMetadata(BaseModel):
     after: Optional[List[FidesKey]]
 
 
-class Dataset(FidesModel, FidesopsMetaBackwardsCompat):
+class Dataset(FidesModel, FidesopsMetaBackwardsCompat):  # type: ignore[misc]
```
ThomasLaPiana: academically i'm curious what the type error is?

ThomasLaPiana (changes requested):
small abstraction change request

src/fideslang/models.py
adamsachs: oh nice, hadn't noticed those - good idea! will go for it

src/fideslang/models.py
adamsachs: complaining about explicit `Any`. it was odd to me that it was complaining both on the line defining the `meta` attribute (L532) and the line defining the wrapping class (L529), but it was...so i just decided to satisfy it 🤷 

i.e. here's what i get from `mypy` when i remove it:
```
src/fideslang/models.py:533: error: Explicit "Any" is not allowed  [misc]
    class Dataset(FidesModel, FidesopsMetaBackwardsCompat):

```

src/fideslang/models.py
adamsachs: updated in https://github.com/ethyca/fideslang/pull/113/commits/070d05cb40272fd0222d587929fd48d417f44888

src/fideslang/models.py
ThomasLaPiana: interesting, thanks for the note!

src/fideslang/models.py
```diff
@@ -922,9 +925,7 @@ class System(FidesModel):
     registry_id: Optional[int] = Field(
         description="The id of the system registry, if used.",
     )
-    meta: Optional[Dict[str, str]] = Field(
-        description="An optional property to store any extra information for a system. Not used by fidesctl.",
-    )
+    meta: Dict = meta_field
```
ThomasLaPiana, line 928: I realized that since we actually want to allow arbitrary stuff, there is no point in type enforcing the object internals! this also removes the need for the ignores

src/fideslang/models.py
```diff
@@ -53,6 +54,10 @@
     default=False,
     description="Denotes whether the resource is part of the default taxonomy or not.",
 )
+meta_field = Field(
+    default_factory=dict,
```
ThomasLaPiana, line 58: I don't see a reason to allow `None` here, this way users can do `dataset.meta.get("somekey")` without first having to do an existence check, and it will still evaluate to false in a bool check

ThomasLaPiana (approved):

PR #121, opened 2023-06-08
Dev UX: allow docker to select platform

ewdurbin:
Closes NA

This is a developer user experience concern. I run on an M1 Mac and this hard-coded platform (for reasons I can't determine) makes working on fideslang locally very slow.

### Code Changes

* [ ] Remove hardcoded --platform from Dockerfile

### Steps to Confirm

* [ ] Run `make check-all` and note that things work

### Pre-Merge Checklist

* [ ] Documentation Updated
* [ ] Update `CHANGELOG.md`

### Description Of Changes

If there's something I'm missing as to why a pure python project needs to be hard pinned to dev/run/build on amd64, this is probably not an acceptable change.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 python:3.8-slim-bullseye as base
```
ThomasLaPiana, line 1: I think this was as copy/pasta from `fides` 😬 nice catch here, I run on a windows machine so I never notice these things!

ThomasLaPiana (approved):

ThomasLaPiana:
I just realized the branch name, hah

PR #135, opened 2023-06-23
Consolidate Python build/configuration into pyproject.toml

### LINKED ISSUE #125: Consolidate build/tool configuration into `pyproject.toml`

Currently spread across `setup.py`, `setup.cfg` and `pyproject.toml`. Modern Python supports keeping everything in `pyproject.toml`

Consolidation

ThomasLaPiana:
Closes #125 

### Code Changes

* [x] move `setup.py` into `pyproject.toml`
* [x] move `setup.cfg` into `pyproject.toml`
* [x] get rid of the old `optional-requirements.txt` (not sure what it is used for)
* [x] replace `versioneer` with `setuptools_scm`
* [x] prep `CHANGELOG` for a release to test these changes

### Steps to Confirm

* [ ] `pip install -e .`
* [ ] `python -c "import fideslang; print(fideslang.__version__)"`

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Relevant Follow-Up Issues Created (there is a super old issue for this over in fides: [issue](https://github.com/ethyca/fides/issues/1219))
* [x] Update `CHANGELOG.md`

### Description Of Changes

This PR has two primary objectives:
1. Modernize our packaging methodology to keep pace with current Python Best Practices
2. Simplify our development and packaging (a single file, cohesive tooling, no gotchas)
3. Get some practice before trying to tackle updating `fides` in the same way

Additionally I'll run a `fideslang` release after this PR merges to confirm everything is working as expected so we don't get "surprised" if something goes wrong on the next release


pyproject.toml
```diff
@@ -1,25 +1,76 @@
...
+where = ["src"]
+
+############
+## Typing ##
+############
+[tool.mypy]
+check_untyped_defs = true
+disallow_untyped_defs = true
+disallow_any_explicit = true
+files = ["src"]
```
pattisdr, line 47: Does this need a backslash after it?

pattisdr (commented):
Nice consolidation into the preferred pyproject.toml! I'm not an expert here, but I did some research and it looks like the new setup sections in the pyproject.toml have been formatted appropriately - I can't say for sure that everything is done correctly here.

pyproject.toml
adamsachs: doesn't look like we've got one in `fides` 🤷

Dockerfile
```diff
@@ -33,9 +33,6 @@ RUN pip install -r dev-requirements.txt
 COPY requirements.txt requirements.txt
 RUN pip install -r requirements.txt
 
-COPY optional-requirements.txt optional-requirements.txt
```
adamsachs, line 36: is this just unrelated cleanup? just want to be sure

data_uses.csv
adamsachs, line 1: also unrelated cleanup? just for clarification

adamsachs (approved):
looks good to me, i'm also not an expert but nothing stood out! just some minor comments for clarification purposes. this is a nice update to streamline and get us on the latest best practice 👍 

we may need to be a _bit_ more careful in fides just because we rely on versioning for tagging (and maybe other things?) - not that anything looks off here, just that we'll wanna be sure we test out anything else that relies on the version number? this is def a nice practice run though, as you mention!

Dockerfile
ThomasLaPiana: yep! I do not at all remember why I had this in there...

data_uses.csv
ThomasLaPiana: yep, I accidentally committed that in another recent PR

ThomasLaPiana:
@adamsachs it'll definitely be trickier, for instance I think the version formatting might break when we try to upload to Test PyPI for instance but we'll cross that bridge when we come to it!

PR #137, opened 2023-07-14
add `active` property to taxonomy classes

adamsachs:
needed for https://github.com/ethyca/fides/issues/3709

### Description Of Changes

Adds an `active` flag to the models for taxonomy elements. Flag defaults to `true`.

### Code Changes

* [x] add `active` boolean property to the taxonomy element models: `DataUse`, `DataSubject`, `DataQualifier`, `DataCategory`

### Steps to Confirm

not much here - corresponding [`fides` PR](https://github.com/ethyca/fides/pull/3784) has some actionable confirmation steps

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ The types of changes are:
 
 ## [Unreleased](https://github.com/ethyca/fideslang/compare/1.4.3...main)
 
+### Added
+
+- Add an `active` boolean property to taxonomy elements [#137](https://github.com/ethyca/fideslang/pull/137)
+
 ### Fixed
 
 - Don't allow duplicate entries for DatasetCollections as part of Datasets [#136](https://github.com/ethyca/fideslang/pull/136)
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -27,8 +27,8 @@ from fideslang.validation import (
     no_self_reference,
     parse_data_type_string,
     sort_list_objects_by_name,
-    valid_data_type,
     unique_items_in_list,
+    valid_data_type,
 )
 
 # Reusable Validators
@@ -55,6 +55,9 @@ is_default_field = Field(
 meta_field = Field(
     default=None,
     description="An optional property to store any extra information for a resource. Data can be structured in any way: simple set of `key: value` pairs or deeply nested objects.",
+)
+active_field = Field(
+    default=True, description="Indicates whether the resource is currently 'active'."
 )
 
 
@@ -164,6 +167,7 @@ class DataCategory(FidesModel):
 
     parent_key: Optional[FidesKey]
     is_default: bool = is_default_field
+    active: bool = active_field
 
     _matching_parent_key: classmethod = matching_parent_key_validator
     _no_self_reference: classmethod = no_self_reference_validator
@@ -174,6 +178,7 @@ class DataQualifier(FidesModel):
 
     parent_key: Optional[FidesKey]
     is_default: bool = is_default_field
+    active: bool = active_field
 
     _matching_parent_key: classmethod = matching_parent_key_validator
     _no_self_reference: classmethod = no_self_reference_validator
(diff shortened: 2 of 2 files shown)


adamsachs:
@ThomasLaPiana just a small `fideslang` update that's needed for some `fides` functionality. once we get this PR merged and the packaging issues i describe in https://github.com/ethyca/fides/pull/3784#issuecomment-1636424081 fixed, i'd like to release a point-release of `fideslang` here so we can get these updates properly into `fides`!

ThomasLaPiana:
@adamsachs I'm not sure I'm onboard with this change, can we discuss here a bit?

I totally understand the need for the core change itself (hiding certain taxonomy items in the UI) but this doesn't feel like the right place to do it.

Why do we need to update `Fideslang` to make it so that certain items don't show up in a UI, in a downstream dependency? It doesn't seem like the `active` flag adds any value to `Fideslang` in isolation. This is a tight coupling of functionality.

As an alternative, could we add this as a database field in `Fides` directly to keep track of? Or somehow store it elsewhere without `Fideslang` needing an update?

I know this is definitely the "fastest" solution, but I'm pushing back here since I don't think it is the right one

adamsachs:
> @adamsachs I'm not sure I'm onboard with this change, can we discuss here a bit?
> 
> I totally understand the need for the core change itself (hiding certain taxonomy items in the UI) but this doesn't feel like the right place to do it.
> 
> Why do we need to update `Fideslang` to make it so that certain items don't show up in a UI, in a downstream dependency? It doesn't seem like the `active` flag adds any value to `Fideslang` in isolation. This is a tight coupling of functionality.
> 
> As an alternative, could we add this as a database field in `Fides` directly to keep track of? Or somehow store it elsewhere without `Fideslang` needing an update?
> 
> I know this is definitely the "fastest" solution, but I'm pushing back here since I don't think it is the right one

@ThomasLaPiana good call out - thank you for pushing back! i get your point - i was a bit hasty here to put down the most straightforward path in terms of implementation, but i think you're right that this creates an unnecessary and undesirable coupling here in `fideslang`.

broadly, the two alternative paths i see for implementation over in `fides` (and we can move this conversation to the issue itself https://github.com/ethyca/fides/issues/3709, since it probably doesn't belong in `fideslang` anymore):
1. implement a "mixin" base model class (something like `SupportsActiveMixin`) that's overridden by ("mixed into") a new pydantic model used for defining the API request/response structures for taxonomy CRUD updates. for an API client, the end result would basically look identical to the updates i'd proposed here, though the implementation would obviously differ on the backend.
2. define and implement a different API pattern for updating the `active` property on our taxonomy elements, e.g. an entirely new endpoint that's geared specifically toward setting the `active` state of one or more taxonomy elements, an endpoint separate from the "basic" CRUD updates to these elements;

of those 2, i'm leaning toward option 1 because i think it keeps a more consistent and easy-to-use API for the client, even if the implementation is perhaps a _bit_ less straightforward on the BE. i'd like to get your POV though - do those two options make sense? if so, any preference? any other ideas that differ from those?

ThomasLaPiana:
@adamsachs is there a reason not to extend the existing Pydantic classes for those models? That seems like the simplest solution to me but maybe I'm missing something there that required the mixin

adamsachs:
> @adamsachs is there a reason not to extend the existing Pydantic classes for those models? That seems like the simplest solution to me but maybe I'm missing something there that required the mixin

oh yeah, i just thought a mixin is a slightly cleaner way of doing that - i.e. extend the models in `fides`, but extend them by sharing a common mixin base class (which defines the `active` field) rather than all of the extensions "happening" to have the same `active` field.

maybe i'm not thinking of mixins correctly though? anyway, i think we're roughly on the same page, so i can get a draft PR up for review and that'll probably be easier than trying to talk through it at this point 😄

adamsachs:
closing as we're going in a different direction per above conversation - see https://github.com/ethyca/fides/pull/3784, the changes are self-contained there in `fides` 👍

Issue #139, opened 2023-07-20
Deprecate legal_basis , legitimate_interest, and legitimate_interest_impact_assessment in taxonomy


rsilvery:
When #3827 is complete, we need to deprecate these fields in fideslang and provide a migration path.

pattisdr:
Reopening, I am not sure if I am deprecating these yet in the same body of work

rsilvery:
We are, paired with #3907 that we'll deliver in this release.

pattisdr:
Fideslang support here: https://github.com/ethyca/fideslang/pull/146, deprecates these fields, and others in System, PrivacyDeclaration, Data Use, and Dataset.

Roger-Ethyca:
moving to done

PR #146, opened 2023-08-09
System + Privacy Declaration Dictionary Changes

### LINKED ISSUE #3833: Backend work to support expanded system forms
Need to determine which backend changes will be needed to support saving fields added in #3827
Also need to store features from the GVL

pattisdr:
Partially closes [#3833](https://github.com/ethyca/fides/issues/3833) and https://github.com/ethyca/fideslang/issues/139

### Description Of Changes

Adds new fields to System and Privacy Declaration to support new dictionary feature.  Deprecates related fields in System/PrivacyDeclaration/DataUse/Datasets.  The goal here is for all of these changes to be additive, and we can remove the deprecated fields at a later time.  

On the Fides side, we are migrating over data from the deprecated fields to their new locations, and hiding these fields in the UI, but leaving the original fields in place for now.

Context is here: [Confluence Dictionary Forms](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2742648890/Dictionary+forms) and [Fideslang Updates](https://ethyca.atlassian.net/wiki/spaces/PM/pages/2740027499/Fideslang+updates)

### Code Changes

* [x] Deprecates the following fields: 
    - Dataset: `data_qualifier`, `joint_controller`, `retention`, and `third_country_transfers`
       - The idea is that these are more legal terms that don't belong on more of a data engineering construct. `data_qualifier` is being removed because we're not using.  The other three are being migrated/transformed to `System.joint_controller_info`, `PrivacyDeclaration.retention_period`, and `System.does_international_transfers`
   - Privacy Declaration: `data_qualifier`
       - Being removed as it's not being used.
    - System: `data_responsibility_title`, `joint_controller`, `third_country_transfers`, `data_protection_impact_assessment`. 
        -  These are being migrated/transformed to `System.responsibility`, `System.joint_controller_info`, `System.does_international_transfers`, `System.requires_data_protection_assessments`, `System.dpa_location`, and `System.dpa_progress`
    -  Data Use: `legal_basis`, `recipients`, `special_category`, `legitimate_interest`, `legitimate_interest_impact_assessment`.  
        - One is being dropped altogether, the other four are being moved/transformed to PrivacyDeclaration
* [x] Adds new fields to System and PrivacyDeclaration to support new Dictionary requirements

### Steps to Confirm

* [ ] See this in action by creating and updating systems on fides branch https://github.com/ethyca/fides/pull/3901, which is pinned to the latest commit here.  

### Pre-Merge Checklist

* [ ] Documentation Updated
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -4,31 +4,19 @@
 Contains all of the Fides resources modeled as Pydantic models.
 """
 from __future__ import annotations
-from warnings import warn
+
 from enum import Enum
 from typing import Any, Dict, List, Optional, Union
+from warnings import warn
 
-from pydantic import (
-    AnyUrl,
-    BaseModel,
-    ConstrainedStr,
-    Field,
-    HttpUrl,
-    PositiveInt,
-    root_validator,
-    validator,
-)
+from pydantic import (AnyUrl, BaseModel, ConstrainedStr, Field, HttpUrl,
+                      PositiveInt, root_validator, validator)
 
-from fideslang.validation import (
-    FidesKey,
-    check_valid_country_code,
-    matching_parent_key,
-    no_self_reference,
-    parse_data_type_string,
-    sort_list_objects_by_name,
-    unique_items_in_list,
-    valid_data_type,
-)
+from fideslang.validation import (FidesKey, check_valid_country_code,
+                                  matching_parent_key, no_self_reference,
+                                  parse_data_type_string,
+                                  sort_list_objects_by_name,
+                                  unique_items_in_list, valid_data_type)
 
 # Reusable Validators
 country_code_validator = validator("third_country_transfers", allow_reuse=True)(
@@ -1011,6 +999,12 @@ class System(FidesModel):
     ingress: Optional[List[DataFlow]] = Field(
         description="The resources from which the System receives data."
     )
+    destination: Optional[List[DataFlow]] = Field(
+        description="The resources to which the System sends data."
+    )
+    source: Optional[List[DataFlow]] = Field(
+        description="The resources from which the System receives data."
+    )
     privacy_declarations: List[PrivacyDeclaration] = Field(
         description=PrivacyDeclaration.__doc__,
     )
@@ -1082,10 +1076,12 @@ class System(FidesModel):
     dpo: Optional[str] = Field(
         description="The official privacy contact address or DPO."
     )
-    joint_controller: Optional[str] = Field(
+    joint_controller: Optional[ContactDetails] = Field(
+        description=ContactDetails.__doc__,
+    )
+    joint_controller_info: Optional[str] = Field(
         description="The party or parties that share the responsibility for processing personal data."
     )
-
     data_security_practices: Optional[str] = Field(
         description="The data security practices employed by this system."
     )
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -9,14 +9,27 @@ from enum import Enum
 from typing import Any, Dict, List, Optional, Union
 from warnings import warn
 
-from pydantic import (AnyUrl, BaseModel, ConstrainedStr, Field, HttpUrl,
-                      PositiveInt, root_validator, validator)
+from pydantic import (
+    AnyUrl,
+    BaseModel,
+    ConstrainedStr,
+    Field,
+    HttpUrl,
+    PositiveInt,
+    root_validator,
+    validator,
+)
 
-from fideslang.validation import (FidesKey, check_valid_country_code,
-                                  matching_parent_key, no_self_reference,
-                                  parse_data_type_string,
-                                  sort_list_objects_by_name,
-                                  unique_items_in_list, valid_data_type)
+from fideslang.validation import (
+    FidesKey,
+    check_valid_country_code,
+    matching_parent_key,
+    no_self_reference,
+    parse_data_type_string,
+    sort_list_objects_by_name,
+    unique_items_in_list,
+    valid_data_type,
+)
 
 # Reusable Validators
 country_code_validator = validator("third_country_transfers", allow_reuse=True)(
@@ -170,8 +183,8 @@ class SpecialCategoriesEnum(str, Enum):
 
 class SpecialCategoryLegalBasisEnum(str, Enum):
     """
-    The model for processing special categories
-    of personal data.
+    The model for the legal basis for processing special categories of personal data
+    on privacy declarations
 
     Based upon article 9 of the GDPR
     """
@@ -1183,7 +1196,7 @@ class System(FidesModel):
         values: Dict,
     ) -> PrivacyDeclaration:
         """
-        Any `PrivacyDeclaration`s which include `source` and/or `destination` fields must
+        Any `PrivacyDeclaration`s which include `egress` and/or `ingress` fields must
         only reference the `fides_key`s of defined `DataFlow`s in said field(s).
         """
 
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -133,7 +133,7 @@ class LegalBasisForProcessingEnum(str, Enum):
 
     CONSENT = "Consent"
     CONTRACT = "Contract"
-    LEGAL_OBLIGATION = "Legal Obligation" # TODO: spec indicates this should now be "Legal claims" - do we update?
+    LEGAL_OBLIGATION = "Legal Obligation"  # TODO: spec indicates this should now be "Legal claims" - do we update?
     VITAL_INTEREST = "Vital Interest"
     PUBLIC_INTEREST = "Public Interest"
     LEGITIMATE_INTEREST = "Legitimate Interests"
@@ -152,7 +152,9 @@ class LegalBasisForTransfersEnum(str, Enum):
     other = "Other"
 
 
-class SpecialCategoriesEnum(str, Enum): # TODO: spec has updated friendly text here - do we need to update?
+class SpecialCategoriesEnum(
+    str, Enum
+):  # TODO: spec has updated friendly text here - do we need to update?
     """
     The model for processing special categories
     of personal data.
@@ -842,7 +844,7 @@ class PrivacyDeclaration(BaseModel):
         description="The fides key of the data qualifier describing a system in a privacy declaration.",
     )
     data_subjects: List[FidesKey] = Field(
-        default=[]
+        default=[],
         description="An array of data subjects describing a system in a privacy declaration.",
     )
     egress: Optional[List[FidesKey]] = Field(
@@ -855,8 +857,7 @@ class PrivacyDeclaration(BaseModel):
         description="Cookies associated with this data use to deliver services and functionality",
     )
     features: List[str] = Field(
-        default=[],
-        description="The features of processing personal data."
+        default=[], description="The features of processing personal data."
     )
     legal_basis_for_processing: Optional[LegalBasisForProcessingEnum] = Field(
         description="The features of processing personal data."
@@ -866,7 +867,7 @@ class PrivacyDeclaration(BaseModel):
     )
     processes_special_category_data: bool = Field(
         default=False,
-        description="The amount of time (in days) for which data is retained for this purpose."
+        description="The amount of time (in days) for which data is retained for this purpose.",
     )
     special_category: Optional[SpecialCategoriesEnum] = Field(
         description="The legal basis under which the special category data is processed.",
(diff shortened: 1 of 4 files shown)


src/fideslang/models.py
```diff
@@ -126,9 +126,7 @@ class DataSubjectRightsEnum(str, Enum):
 
 class LegalBasisEnum(str, Enum):
     """
-    The model for allowable legal basis categories
-
-    Based upon article 6 of the GDPR
+    Deprecated. The model for allowable legal basis categories on data uses.
```
pattisdr, line 129: This Enum is for a field on Data Uses, but that field will soon be going away.  A new version with different casing has been added: LegalBasisForProcessingEnum for use on the similar field in PrivacyDeclaration.    Adding a new one to not migrate data on the existing column.

When we deprecate DataUse.legal_basis, we can get rid of this enum here.

src/fideslang/models.py
```diff
@@ -139,12 +137,41 @@ class LegalBasisEnum(str, Enum):
...
+class LegalBasisForTransfersEnum(str, Enum):
+    """The model for describing the legal basis under which data is transferred"""
+
+    ADEQUACY_DECISION = "Adequacy decision"
+    STANDARD_CONTRACTUAL_CLAUSES = "Standard contractual clauses"
+    BINDING_CORPORATE_RULES = "Binding corporate rules"
+    OTHER = "Other"
+
+
+class SpecialCategoriesEnum(str, Enum):
```
pattisdr, line 172: Similar, this enum is used on a field that is now deprecated, so we can remove when DataUse.special_category is removed.  I've created a new enum for SpecialCategoryLegalBasisEnum which has different options and casing to be used for the new field in PrivacyDeclaration.

src/fideslang/models.py
```diff
@@ -839,20 +929,49 @@ class PrivacyDeclaration(BaseModel):
...
-        Warn that the `dataset_references` field is deprecated, if set.
+        Warn that the `data_qualifier` field is deprecated, if set.
         """
-
         if value is not None:
             warn(
-                "The dataset_references field is deprecated, and will be removed in a future version of fideslang. Use the 'egress' and 'ingress` fields instead.",
+                "The data_qualifier field is deprecated, and will be removed in a future version of fideslang.",
                 DeprecationWarning,
             )
```
pattisdr, line 969: Note that from a discussion w/ Neville, Rachel, Kelsey, and Adrian on 8/8, we are removing the deprecation of PrivacyDeclaration.dataset_references.  This field has been marked in fideslang as having been deprecated for awhile, but we've been continuing to write to this field in the UI.  This field still seems to be useful, and in fact, we're adding this field at the System level as well, so it will be used in both places.

I've been asked to get rid of the deprecation, so I'm just swapping it with the new data_qualifier deprecation here.

src/fideslang/models.py
```diff
@@ -970,25 +1089,109 @@ class System(FidesModel):
...
+    responsibility: List[DataResponsibilityTitle] = Field(
+        default=[],
+        description=DataResponsibilityTitle.__doc__,
+    )
+    dpo: Optional[str] = Field(
+        description="The official privacy contact address or DPO."
+    )
+    joint_controller_info: Optional[str] = Field(
+        description="The party or parties that share the responsibility for processing personal data."
+    )  # Use joint_controller_info in favor of joint_controller
```
pattisdr, line 1155: In the spirit of trying to reduce risk with so many fields moving around, I'm trying to keep everything additive here, and have added a new `joint_controller_info` field.  `System.joint_controller` is a dictionary format (that is stored encrypted in our database) and the new format will be a string that is not encrypted.  This felt like enough of a change to warrant adding a new field and deprecating the previous.  Another reason for this is that we're actually combining information from `System.joint_controller` and `Datasets.joint_controller` in the data migration.

src/fideslang/models.py
```diff
@@ -839,20 +929,49 @@ class PrivacyDeclaration(BaseModel):
...
+    retention_period: Optional[str] = Field(
+        description="An optional string to describe the time period for which data is retained for this purpose."
+    )
+    processes_special_category_data: bool = Field(
+        default=False,
+        description="This system processes special category data",
+    )
+    special_category_legal_basis: Optional[SpecialCategoryLegalBasisEnum] = Field(
+        description="The legal basis under which the special category data is processed.",
+    )
```
pattisdr, line 942: I've held off on adding any extra validation here but let me know if you disagree.  

For example, if processes_special_category_data is True, we should likely have "special_category_legal_basis_ populated and so on.  There are many more similar examples to this, but it seems like that might too tightly constrain the customer.  They may not know the special category legal basis yet, so this still allows them to populate processes_special_category_data is True, to fill out what they can.

src/fideslang/models.py
```diff
@@ -239,29 +286,51 @@ class DataUse(FidesModel):
...
+        """
+        deprecated_fields = [
+            "legal_basis",
+            "recipients",
+            "special_category",
+            "legitimate_interest",
+            "legitimate_interest_impact_assessment",
+        ]
+        for field in deprecated_fields:
+            if values.get(field) is not None:
```
ThomasLaPiana, line 322: Nit: `None` is the default fallback value for `.get`

```suggestion
            if values.get(field):
```

src/fideslang/models.py
```diff
@@ -839,20 +929,49 @@ class PrivacyDeclaration(BaseModel):
     ingress: Optional[List[FidesKey]] = Field(
         description="The resources from which data is received. Any `fides_key`s included in this list reference `DataFlow` entries in the `ingress` array of any `System` resources to which this `PrivacyDeclaration` is applied."
     )
+    features: List[str] = Field(
+        default=[], description="The features of processing personal data."
```
ThomasLaPiana: Default values like this are dangerous in Python and can lead to some nasty unintended side-effects (been there, felt that pain!)

```suggestion
        default_factory=list, description="The features of processing personal data."
```

src/fideslang/models.py
```diff
@@ -839,20 +929,49 @@ class PrivacyDeclaration(BaseModel):
...
+    )
+    data_shared_with_third_parties: Optional[bool] = Field(
+        default=False,
+        description="This system shares data with third parties for this purpose.",
+    )
+    third_parties: Optional[str] = Field(
+        description="The types of third parties the data is shared with.",
+    )
+    shared_categories: List[str] = Field(
+        default=[],
```
ThomasLaPiana: Same as above

```suggestion
        default_factory=list,
```

src/fideslang/models.py
```diff
@@ -839,20 +929,49 @@ class PrivacyDeclaration(BaseModel):
...
+        description="An optional string to describe the time period for which data is retained for this purpose."
+    )
+    processes_special_category_data: bool = Field(
+        default=False,
+        description="This system processes special category data",
+    )
+    special_category_legal_basis: Optional[SpecialCategoryLegalBasisEnum] = Field(
+        description="The legal basis under which the special category data is processed.",
+    )
+    data_shared_with_third_parties: Optional[bool] = Field(
```
ThomasLaPiana: If we're already defaulting to `False` I don't think the `Optional` is required here?

```suggestion
    data_shared_with_third_parties: bool = Field(
```

src/fideslang/models.py
ThomasLaPiana: Totally agree here! In the end if it's going to be better for our customers, it's probably the right choice 🙂

src/fideslang/models.py
ThomasLaPiana: Excellent, sounds good!

src/fideslang/models.py
```diff
@@ -970,25 +1089,109 @@ class System(FidesModel):
...
+        description="A URL that points to the System's publicly accessible privacy policy."
+    )
+    legal_name: Optional[str] = Field(
+        description="The legal name for the business represented by the system."
+    )
+    legal_address: Optional[str] = Field(
+        description="The legal address for the business represented by the system."
+    )
+    responsibility: List[DataResponsibilityTitle] = Field(
+        default=[],
```
ThomasLaPiana:
```suggestion
        default_factory=list,
```

tests/fideslang/test_models.py
```diff
@@ -390,3 +543,69 @@ def test_valid_dataset(self):
                 ),
             ],
         )
+
+    def test_dataset_data_qualifier_deprecation(self) -> None:
```
ThomasLaPiana: I'm not sure what the efficacy of testing every single deprecated value is here. It seems like if one works, they'd all work?

Additionally, given the similarities here, I would rather see a single parametrized test instead of multiple discrete tests

src/fideslang/models.py
```diff
@@ -970,25 +1089,109 @@ class System(FidesModel):
...
+        Warn of System fields pending deprecation.
+        """
+        deprecated_fields = [
+            "joint_controller",
+            "third_country_transfers",
+            "data_responsibility_title",
+            "data_protection_impact_assessment",
+        ]
+        for field in deprecated_fields:
+            if values.get(field) is not None:
```
ThomasLaPiana, line 1179:
```suggestion
            if values.get(field):
```

ThomasLaPiana (commented):
Huge work here! I left some nits but structurally looks great

pattisdr:
Thanks very much for all your comments @ThomasLaPiana!

src/fideslang/models.py
pattisdr: My intent here was to show a warning if anything at all was passed in that wasn't null -

src/fideslang/models.py
```diff
@@ -823,11 +912,12 @@ class PrivacyDeclaration(BaseModel):
...
         description="The Data Use describing a system in a privacy declaration.",
     )
-    data_qualifier: FidesKey = Field(
+    data_qualifier: Optional[FidesKey] = Field(
         default="aggregated.anonymized.unlinked_pseudonymized.pseudonymized.identified",
-        description="The fides key of the data qualifier describing a system in a privacy declaration.",
+        description="Deprecated. The fides key of the data qualifier describing a system in a privacy declaration.",
     )
     data_subjects: List[FidesKey] = Field(
+        default=[],
```
TheAndrewJackson: Same suggestion as @ThomasLaPiana below

```suggestion
        default_factory=list,
```

src/fideslang/models.py
pattisdr: I'll make the change, but my understanding was that the using a default=[] or `default_factory`=list results in the same behavior though, given that Pydantic deepcopies the default value for each new instance  https://docs.pydantic.dev/latest/usage/models/#fields-with-non-hashable-default-values

Can you clarify how this will make a difference?

src/fideslang/models.py
pattisdr: yep thanks!

tests/fideslang/test_models.py
pattisdr: I can consolidate! My intent wasn't to test that it works, rather to add an additional layer in the tests protecting which fields are considered deprecated.

src/fideslang/models.py
pattisdr: Example from one of my tests - if DataUse.legitimate_interest is set to False, if we use `values.get(field)` there is no deprecation warning thrown, but I think there should be.  They've set a value.

pattisdr: This made me think of something else - I feel I should remove default values from these deprecated fields.  Make sure they're all optional and no defaults. Otherwise, they'll be getting warnings even if they're not supplying values for deprecated fields.

src/fideslang/models.py
pattisdr: thanks responded above! https://github.com/ethyca/fideslang/pull/146#discussion_r1290219668

pattisdr:
Back to you @ThomasLaPiana and @TheAndrewJackson thanks again for looking at my PR!

src/fideslang/models.py
ThomasLaPiana: You're totally right, I was mistaken! It looks like Pydantic fixes Python's typical unexpected behavior here 🙂 

TIL!

src/fideslang/models.py
ThomasLaPiana: Makes sense 🤔

ThomasLaPiana (approved):
I see you still want to make another adjustment to default values for deprecated fields, but once you're satisfied feel free to ship it!

pattisdr:
They're included here too @ThomasLaPiana!  Thanks so much for taking another look ⭐

PR #147, opened 2023-08-11
Add version information to Default Taxonomy models

### LINKED ISSUE #129: [Epic] Add the ability to gracefully deprecate parts of the default taxonomy
# Problem Statement

Given the ever-evolving nature of the privacy landscape, we need to continually evolve the default taxonomy (provided by `fideslang`). While it is easy to make changes to the taxonomy itself, the downstream effects of needing to migrate are significantly more complex.

# Proposed Solution

For all intents and purposes, the `Default Taxonomy` should be considered a [slowly changing dimension](https://en.wikipedia.org/wiki/Slowly_changing_dimension) (data warehousing term) and as such we should employ a common technique for handling slowly changing dimensions.

To be able to preserve unlimited history, I support the [add row method](https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row).

By effectively adding canonical versioning to the `Default Taxonomy` itself (not just based on pip version numbers) we can significantly simplify

### Potential Dangers

By basically now promising to keep infinite history, this would complicate potentially major changes to the default taxonomy models. For instance, if we were to add or remove fields from those models we'd need to make sure that we handled it in a backwards compatible way, or update the deprecated models as well.

This would mean more maintenance over time, but I believe this cost is worth the trade-off in overall stability and usability.

#### Example 

```python
class DataSubject(FidesModel):
    """The DataSubject resource model."""

    rights: Optional[DataSubjectRights] = Field(description=DataSubjectRights.__doc__)
    automated_decisions_or_profiling: Optional[bool] = Field(
        description="A boolean value to annotate whether or not automated decisions/profiling exists for the data subject.",
    )
    is_default: bool = is_default_field
    version_introduced: Version
    version_deprecated: Version # This should be a validator to ensure we are using correct semver formatting
    replaced_by: FidesKey
```

ThomasLaPiana:
Closes #129 

### Description Of Changes

The idea here is to add fields to the default taxonomy that enable us to programmatically version each record.

This is useful for doing things like data migrations over in `Fides` as well as signaling to users when something gets deprecated without removing it in a totally destructive manner.

### Code Changes

* [x] add the new fields to the taxonomy models
* [x] Refactor default taxonomy to use a factory for default items, as well as store the new fields in a separate class for inheritance
* [x] add validators and tests for the new functionality

### Steps to Confirm

* [ ] new tests should cover everything

### Pre-Merge Checklist

* [x] Documentation Updated
* [x] Update `CHANGELOG.md`


src/fideslang/models.py
```diff
@@ -52,15 +64,26 @@
     default=False,
     description="Denotes whether the resource is part of the default taxonomy or not.",
 )
+version_added_field = Field(
+    default=None,
+    description="The version of Fideslang in which this record was added.",
+)
+version_deprecated_field = Field(
+    default=None,
+    description="The version of Fideslang in which this record was deprecated.",
```
ThomasLaPiana: Not sure on wording here...`record` feelings wrong but I'm not sure what else to use. Object?

src/fideslang/default_taxonomy/utils.py
```diff
@@ -0,0 +1,22 @@
+from typing import Dict, Union
+
+from fideslang.models import DataCategory, DataQualifier, DataSubject, DataUse
+
+CustomType = Union[DataCategory, DataSubject, DataQualifier, DataUse]
+
+
+def default_factory(taxonomy_class: CustomType, **kwargs: Dict) -> CustomType:
```
pattisdr, line 8: Nice implementation here 👍

src/fideslang/models.py
pattisdr: I don't know, entity? construct? item?

src/fideslang/models.py
```diff
@@ -52,15 +64,26 @@
...
+)
+replaced_by_field = Field(
+    default=None,
+    description="The new name, if applicable, for this field after deprecation.",
+)
```
pattisdr: Now that is_default_field is only being used in the DefaultModel, it seems like that and these other new fields could just be defined directly in DefaultModel to make that model easier to parse

src/fideslang/validation.py
```diff
@@ -77,6 +90,58 @@ def no_self_reference(value: FidesKey, values: Dict) -> FidesKey:
...
+        try:
+            assert not values.get("version_added")
+            assert not values.get("version_deprecated")
+            assert not values.get("replaced_by")
+        except AssertionError:
+            raise FidesValidationError(
+                "Non-default items can't have version information!"
+            )
+    # If it's a default item, it at least needs a starting version
+    else:
```
pattisdr: in this case I'd do the affirmative first, "if value; else", I just think it's easier to parse

src/fideslang/validation.py
```diff
@@ -77,6 +90,58 @@ def no_self_reference(value: FidesKey, values: Dict) -> FidesKey:
...
+
+def is_deprecated_if_replaced(value: str, values: Dict) -> str:
+    """
+    Check to make sure that the item has been deprecated if there is a replacement.
+    """
+
+    if value and not values.get("version_deprecated"):
+        raise FidesValidationError("Cannot be replaced without deprecation!")
+
+    return value
```
pattisdr: This generic way of defining these new validators just feels a little less readable to me given that it is the value of the "replaced_by" field that matters here, for example, but isn't immediately clear.

src/fideslang/validation.py
```diff
@@ -77,6 +90,58 @@ def no_self_reference(value: FidesKey, values: Dict) -> FidesKey:
...
+def deprecated_version_later_than_added(
+    value: FidesVersion, values: Dict
+) -> FidesVersion:
+    """
+    Check to make sure that the deprecated version is later than the added version.
+
+    This will also catch errors where the deprecated version is defined but the added
+    version is empty.
+    """
+    if value < values.get("version_added", Version("0")):
```
pattisdr: Won't throw an error if they're added and deprecated in the same version -

pattisdr (approved):
Most comments are personal opinions. I like your implementation here; thanks for taking the lead on this!

src/fideslang/models.py
ThomasLaPiana: very good call!

src/fideslang/validation.py
ThomasLaPiana: good point, I struggled with that while writing them too even.

I'll reassign to a variable that makes it more clear

src/fideslang/validation.py
ThomasLaPiana: I knew I would've missed one! Thank you

src/fideslang/models.py
ThomasLaPiana: in the `README` they're often referred to as `labels` so I went with that 🙂

src/fideslang/validation.py
ThomasLaPiana: good point, feels more intuitive

ThomasLaPiana:
holding off on merge until the upstream gets merged in to keep the PRs cleaner

PR #151, opened 2023-08-17
Convert "Version" types to "str" outside of validation

ThomasLaPiana:
Closes #<issue>

### Description Of Changes

The clever (not so much apparently) `Version` type I created is causing issues with database insertion over in Fides ([PR}())

This uses the version during validation but otherwise treats it as a string. This will be released immediately as `Fideslang` version `2.0.1`

### Code Changes

* [x] add version validation while storing as strings
* [x] add new tests to validate empty values on instantiation

### Steps to Confirm

### Pre-Merge Checklist

* [x] Update `CHANGELOG.md`

### DIFF
--- a/scripts/export_default_taxonomy.py
+++ b/scripts/export_default_taxonomy.py
@@ -30,16 +30,6 @@ def export_yaml() -> None:
         print(f"> Writing YAML to {output_filename}")
         resources = [x.dict() for x in getattr(DEFAULT_TAXONOMY, resource_type)]
 
-        # Clean up specific data types
-
-        resources = [
-            {
-                key: str(value) if isinstance(value, Version) else value
-                for key, value in resource.items()
-            }
-            for resource in resources
-        ]
-
         write_manifest(
             output_filename,
             manifest=resources,
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -91,7 +91,7 @@ class DefaultModel(BaseModel):
     A model meant to be inherited by versioned parts of the Default Taxonomy.
     """
 
-    version_added: Optional[FidesVersion] = Field(
+    version_added: Optional[str] = Field(
         default=None,
         description="The version of Fideslang in which this label was added.",
     )
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -116,19 +116,29 @@ class DefaultModel(BaseModel):
 
     @validator("version_added")
     @classmethod
-    def validate_verion_added(cls, version_added: str, values: Dict) -> str:
+    def validate_verion_added(
+        cls, version_added: Optional[str], values: Dict
+    ) -> Optional[str]:
         """
         Validate that the `version_added` field is a proper FidesVersion
         """
+        if not version_added:
+            return None
+
         FidesVersion.validate(version_added)
         return version_added
 
     @validator("version_deprecated")
     @classmethod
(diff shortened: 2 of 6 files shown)


scripts/export_default_taxonomy.py
```diff
@@ -30,16 +30,6 @@ def export_yaml() -> None:
         print(f"> Writing YAML to {output_filename}")
         resources = [x.dict() for x in getattr(DEFAULT_TAXONOMY, resource_type)]
 
-        # Clean up specific data types
-
-        resources = [
```
ThomasLaPiana, line 35: this workaround is no longer needed

src/fideslang/models.py
```diff
@@ -114,6 +114,24 @@ class DefaultModel(BaseModel):
     )
     _is_deprecated_if_replaced: classmethod = is_deprecated_if_replaced_validator
 
+    @validator("version_added")
+    @classmethod
+    def validate_verion_added(cls, version_added: str, values: Dict) -> str:
```
ThomasLaPiana: we want to make sure versions are still proper versions

src/fideslang/models.py
```diff
@@ -91,11 +91,11 @@ class DefaultModel(BaseModel):
     A model meant to be inherited by versioned parts of the Default Taxonomy.
     """
 
-    version_added: Optional[FidesVersion] = Field(
+    version_added: Optional[str] = Field(
         default=None,
         description="The version of Fideslang in which this label was added.",
     )
-    version_deprecated: Optional[FidesVersion] = Field(
+    version_deprecated: Optional[str] = Field(
```
ThomasLaPiana, line 98: this was breaking SQLAlchemy

tests/fideslang/test_validation.py
```diff
@@ -84,6 +84,36 @@ def test_deprecated_after_added(self, TaxonomyClass):
                 version_deprecated="0.2",
             )
 
+    @pytest.mark.parametrize("TaxonomyClass", DEFAULT_TAXONOMY_CLASSES)
+    def test_built_from_dict_with_empty_versions(self, TaxonomyClass) -> None:
+        """Item can't be deprecated in a version earlier than it was added."""
```
ThomasLaPiana, line 89:
```suggestion
        """Test that explicitly building with `None` values is ok."""
```

src/fideslang/models.py
```diff
@@ -114,6 +114,34 @@ class DefaultModel(BaseModel):
     )
     _is_deprecated_if_replaced: classmethod = is_deprecated_if_replaced_validator
 
+    @validator("version_added")
+    @classmethod
+    def validate_verion_added(
```
adamsachs, line 119: couldn't you combine this into a single validation function and annotate to apply it to both fields? i suppose leaving it as two separate validators could improve readability in some senses.

anyway, with only two fields that it applies to, it's certainly easy enough to follow as-is - so definitely not something worth holding up the PR for! just couldn't help but wonder 👍

adamsachs (approved):
lgtm! 👍 just a small comment for curiosity's sake, not suggesting you need to make the update here.

src/fideslang/models.py
ThomasLaPiana: totally true, in this case I decided to just be overly verbose and not worry about over-optimizing. Could definitely improve in the future 😄

PR #160, opened 2023-09-07
Bump Pydantic to version 2

### LINKED ISSUE #159: Update to Pydantic 2.X

There are some big performance improvements in Pydantic 2, that we'll want to take advantage of with our FastAPI server

Will need to be careful see and see how bad any potential breaks might be, although recently moving to `2.0` should make this a bit less painful

ThomasLaPiana:
Closes #159 

### Description Of Changes

Pydantic 2 sees major changes that bring large performance advantages. Worth upgrading to stay up-to-date as well as enjoy downstream effects in increased performance

Pydantic 1 -> 2 Migration Guide: [link](https://docs.pydantic.dev/latest/migration/#migration-guide)

I fully expected we might need to tweak more things here based on testing against `Fides` directly, but we'll need to do more checks there to verify. Here is the [corresponding Fides PR](https://github.com/ethyca/fides/pull/4442)

As part of this migration, I adopted a very different general pattern for validation than we used before. Due to the major changes, I found it easier to use `model_validator` (formerly known as `root_validator`) and composed multiple validators together into a single, full-model validator instead of always using discrete validators by field.

### Code Changes

* [x] make updates required for `Pydantic` v2
* [x] update the requirement bounds in `requirements.txt`
* [x] remove Python 3.8 from supported versions

### Steps to Confirm

* [ ] CI checks pass

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`


requirements.txt
```diff
@@ -1,3 +1,3 @@
-pydantic>=1.8.1,<1.11.0
+pydantic>=2.0.3,<=2.3.0
```
TheAndrewJackson: I'm dropping by for a little drive by engineering here. Pydantic has been updated to `2.4.2` since this PR was first open and `2.5.0` is supposed to come out soon: https://github.com/pydantic/pydantic/milestone/8

Idk if it's worth delaying this for `2.5.0` but it _might_ be worth it depending on the timing. The next version has a lot of bug fixes and performance improvements: https://github.com/pydantic/pydantic/releases/tag/v2.5.0b1

ThomasLaPiana:
@NevilleS a major question we need to answer is if we also want to strip out everything that is `deprecated` as part of 3.0

I think it makes sense to me from a versioning standpoint, but I also want to get more informed opinions from

@SteveDMurphy @pattisdr & @adamsachs

requirements.txt
ThomasLaPiana: @TheAndrewJackson interesting note!

Given that we're trying to make this a library and usable across multiple versions, we'll test across multiple versions and default `Fides` to whatever the newest one is

TheAndrewJackson: That makes sense! It looks like they released `2.5.0` roughly an hour after I typed that comment 😆: https://github.com/pydantic/pydantic/releases/tag/v2.5.0

ThomasLaPiana: They did indeed! Added to the matrix :)

.github/workflows/pr_checks.yml
```diff
@@ -93,9 +93,9 @@ jobs:
   Pytest-Matrix:
     strategy:
       matrix:
-        python_version: ["3.8", "3.9", "3.10", "3.11"]
```
ThomasLaPiana, line 96: Python 3.8 wasn't playing nicely, so I axed it. It is almost EOL and we can only make the matrix so large

.github/workflows/pr_checks.yml
```diff
@@ -93,9 +93,9 @@ jobs:
...
-        pydantic_version: ["1.8.2", "1.9.2", "1.10.9"]
-        pyyaml_version: ["5.4.1", "6.0"]
+        python_version: ["3.9", "3.10", "3.11"]
+        pydantic_version: ["2.2.1", "2.3.0", "2.4.2", "2.5.0"]
```
ThomasLaPiana, line 97: These are the most recent ones, but Pydantic moves fast so we'll need to keep an eye on this

requirements.txt
```diff
@@ -1,3 +1,3 @@
-pydantic>=1.8.1,<1.11.0
+pydantic>=2.2.1,<=2.6.0
```
ThomasLaPiana, line 1: not cross-supporting Pydantic 1.0!

src/fideslang/default_taxonomy/utils.py
```diff
@@ -18,5 +18,5 @@ def default_factory(taxonomy_class: CustomType, **kwargs: Dict) -> CustomType:
         # This is the version where we started tracking from, so
         # we use it as the default starting point.
         kwargs["version_added"] = "2.0.0"  # type: ignore[assignment]
-    item = taxonomy_class.parse_obj(kwargs)
+    item = taxonomy_class.model_validate(obj=kwargs)
```
ThomasLaPiana, line 21: one of the syntax changes

src/fideslang/validation.py
```diff
@@ -3,50 +3,54 @@
...
-        """Validates that the provided string is a valid Semantic Version."""
-        return Version(value)
+    return value
 
 
-class FidesKey(ConstrainedStr):
-    """
-    A FidesKey type that creates a custom constrained string.
-    """
+FidesKey = Annotated[str, PlainValidator(validate_fides_key)]
```
ThomasLaPiana: CustomTypes are now handled very differently, as Annotated instead of Classes that inherit

src/fideslang/validation.py
```diff
@@ -77,91 +81,77 @@ def unique_items_in_list(values: List) -> List:
...
-
-    i.e. DataCategory.parent_key != DataCategory.fides_key
-    """
-    fides_key = FidesKey.validate(values.get("fides_key", ""))
-    if value == fides_key:
-        raise FidesValidationError("FidesKey can not self-reference!")
-    return value
-
-
 def deprecated_version_later_than_added(
```
ThomasLaPiana, line 84: I updated almost all of these functions to be pure functions, not set up as specific validators. They are then more transparently composed on the model itself

tests/conftest.py
```diff
@@ -16,14 +16,14 @@ def resources_dict():
     """
     resources_dict: Dict[str, Any] = {
         "data_category": models.DataCategory(
-            organization_fides_key=1,
+            organization_fides_key="1",
```
ThomasLaPiana, line 19: Pydantic is more picky now! It will not do these subtle conversions (int -> str in this case) but instead will throw an error

tests/fideslang/test_models.py
```diff
@@ -221,71 +217,11 @@ def test_system_valid_no_egress_or_ingress(self) -> None:
                     name="declaration-name",
                 )
             ],
-            registry_id=1,
+            registry_id="1",
             system_type="SYSTEM",
             tags=["some", "tags"],
         )
 
-    def test_system_no_egress(self) -> None:
```
ThomasLaPiana, line 229: I'm confused here, and I think this was a bug? Not sure why these were expected to fail before but aren't now. Especially given that they seem fine given how the model was written? This is really old code though so I'm not sure if just deleting it is right, but I'm not sure what it is solving either. 

I would appreciate any possible insight here!

tests/fideslang/test_validation.py
```diff
@@ -22,7 +21,11 @@
     PrivacyRule,
     System,
 )
-from fideslang.validation import FidesKey, FidesValidationError, valid_data_type
+from fideslang.validation import (
+    FidesValidationError,
+    validate_fides_key,
```
ThomasLaPiana: there is now a separate function for validating the FidesKey since the old method doesn't exist

This is only required when wanting to validate a `FidesKey` outside of a model

tests/fideslang/test_validation.py
```diff
@@ -245,139 +248,120 @@ def test_dataset_duplicate_collections_error():
...
-        name="Custom Test Data",
-        description="Custom Test Data Category",
-    )
-    assert DataCategory
-
-
-@pytest.mark.unit
-def test_fides_key_doesnt_match_stated_parent_key():
-    with pytest.raises(ValidationError):
+class TestFidesKeyValidation:
```
ThomasLaPiana, line 233: Nasty diff here, but it's caused by indenting and making these a test class

src/fideslang/models.py
```diff
@@ -36,29 +38,14 @@
 )
 
 # Reusable Validators
-country_code_validator = validator("third_country_transfers", allow_reuse=True)(
+country_code_validator = field_validator("third_country_transfers")(
     check_valid_country_code
 )
-matching_parent_key_validator = validator("parent_key", allow_reuse=True, always=True)(
```
ThomasLaPiana, line 42: These all got moved and composed into root validators on each model. This pattern doesn't seem to work with Pydantic 2.0 due to the massive changes to validators. This was kind of a hack anyway so I'm not surprised it broke

src/fideslang/models.py
```diff
@@ -280,23 +270,27 @@ class SpecialCategoryLegalBasisEnum(str, Enum):
...
 
     name: str
-    path: Optional[str]
-    domain: Optional[str]
-
-    class Config:
-        """Config for the cookies"""
-
-        orm_mode = True
+    path: Optional[str] = None
```
ThomasLaPiana, line 289: If a field is optional, it requires a default value of `None` now, I guess to be more explicit

src/fideslang/models.py
```diff
@@ -550,47 +575,26 @@ def validate_object_fields(  # type: ignore
...
             )
 
         return fields
 
 
 # this is required for the recursive reference in the pydantic model:
-DatasetField.update_forward_refs()
-
-
-class FidesCollectionKey(ConstrainedStr):
```
ThomasLaPiana, line 569: moved this to `validation.py` and updated it to be the new format

src/fideslang/models.py
```diff
@@ -550,47 +575,26 @@ def validate_object_fields(  # type: ignore
...
-        """
-        values = value.split(".")
-        if len(values) == 2:
-            FidesKey.validate(values[0])
-            FidesKey.validate(values[1])
-            return value
-        raise ValueError(
-            "FidesCollection must be specified in the form 'FidesKey.FidesKey'"
-        )
+DatasetField.model_rebuild()
```
ThomasLaPiana, line 567: this is the new way to do `update_forward_refs()`

src/fideslang/models.py
```diff
@@ -767,11 +770,7 @@ class Evaluation(BaseModel):
         default="",
         description="A human-readable string response for the evaluation.",
     )
-
-    class Config:
-        "Config for the Evaluation"
-        extra = "ignore"
-        orm_mode = True
+    model_config = ConfigDict(extra="ignore", from_attributes=True)
```
ThomasLaPiana, line 729: another change, this time with how models are configured

src/fideslang/models.py
```diff
@@ -937,68 +939,76 @@ class PrivacyDeclaration(BaseModel):
...
     data_categories: List[FidesKey] = Field(
+        default=[],
         description="An array of data categories describing a system in a privacy declaration.",
     )
     data_use: FidesKey = Field(
         description="The Data Use describing a system in a privacy declaration.",
     )
     data_subjects: List[FidesKey] = Field(
-        default_factory=list,
+        default=[],
```
ThomasLaPiana, line 911: I nitted myself here! As taught by @pattisdr it is safe to use `[]` as a default value in Pydantic

pattisdr:
Starting review..a lot to catch up on here!

ThomasLaPiana:
> Starting review..a lot to catch up on here!

I apologize! I just wasn't sure how  to break it up further

I added comments everywhere I thought made sense but of course feel free to dig in and ask other questions and I'll reply as best as I can!

pattisdr:
Oh no this is exciting! Lots of great improvements here, I more meant big picture understanding what Pydantic 2 brings as well.

pattisdr:
There's a whole new lingo to learn!

src/fideslang/models.py
pattisdr: Nice, easier to follow too on the model!

src/fideslang/models.py
```diff
@@ -77,13 +64,13 @@ class FidesModel(BaseModel):
...
-    class Config:
-        "Config for the FidesModel"
-        extra = "ignore"
-        orm_mode = True
+    name: Optional[str] = Field(
+        default=None, description="Human-Readable name for this resource."
+    )
+    description: Optional[str] = Field(
+        default=None, description="A detailed description of what this resource is."
+    )
```
pattisdr, line 70: What was behind pulling out the usage of name_field and description_field here, when you're still using them on other models? Just curious

src/fideslang/models.py
pattisdr: Ah got it thanks!

src/fideslang/models.py
```diff
@@ -1084,48 +1096,60 @@ class System(FidesModel):
...
     administrating_department: Optional[str] = Field(
         default="Not defined",
         description="An optional value to identify the owning department or group of the system within your organization",
     )
     data_protection_impact_assessment: Optional[DataProtectionImpactAssessment] = Field(
-        description="Deprecated. " + DataProtectionImpactAssessment.__doc__,
+        default=None,
+        description="Deprecated. "
+        + (
+            DataProtectionImpactAssessment.__doc__ or ""
```
pattisdr, line 1100: It looks like DataProtectionImpactAssessment > progress/link files might need default=None too (although this is supposed to be deprecated anyway)

src/fideslang/models.py
```diff
@@ -383,21 +394,25 @@ def deprecate_fields(cls, values: Dict) -> Dict:
...
+    @field_validator("legitimate_interest")
     @classmethod
-    def set_legitimate_interest(cls, value: bool, values: Dict) -> bool:
+    def set_legitimate_interest(cls, value: bool, info: ValidationInfo) -> bool:
         """Sets if a legitimate interest is used."""
+        values = info.data
+
         if values["legal_basis"] == "Legitimate Interests":
             value = True
         return value
```
pattisdr: This doesn't seem to run if legitimate_interest is not an argument when instantiating a DataUse. Looks like you can add `validate_default=True` to the `legitimate_interest` Field above to get the same `always=True` behavior.

(Doesn't matter much in practice, this field is deprecated, an argument to remove these deprecated fields soon)

src/fideslang/models.py
```diff
@@ -383,21 +394,25 @@ def deprecate_fields(cls, values: Dict) -> Dict:
...
 
-    @validator("legitimate_interest_impact_assessment", always=True)
+    @field_validator("legitimate_interest_impact_assessment")
     @classmethod
-    def ensure_impact_assessment(cls, value: AnyUrl, values: Dict) -> AnyUrl:
+    def ensure_impact_assessment(cls, value: AnyUrl, info: ValidationInfo) -> AnyUrl:
```
pattisdr: Assume same issue here

src/fideslang/models.py
```diff
@@ -550,47 +575,26 @@ def validate_object_fields(  # type: ignore
...
             data_type, _ = parse_data_type_string(declared_data_type)
             if data_type != "object":
-                raise ValueError(
+                raise FidesValidationError(
                     f"The data type '{data_type}' on field '{field_name}' is not compatible with specified sub-fields. Convert to an 'object' field."
                 )
 
         if (fields or declared_data_type == "object") and values.get("data_categories"):
-            raise ValueError(
+            raise FidesValidationError(
```
pattisdr, line 559: I see, looks like Fideslang is generally throwing a FidesValidationError (which wraps ValueError) instead of ValueError, so this looks to be just updating to match that convention. (I believe I'm the one that brought this ops code over in the first place). 

Just for my own understanding, why is that the convention here?

src/fideslang/models.py
```diff
@@ -1048,22 +1061,21 @@ class DataFlow(BaseModel):
         description="An array of data categories describing the data in transit.",
     )
 
-    @root_validator(skip_on_failure=True)
```
pattisdr, line 1052: What was the `skip_on_failure=True` flag doing here? It looks like with tests, it wasn't skipping on failure originally anyway?

src/fideslang/models.py
```diff
@@ -1084,48 +1096,60 @@ class System(FidesModel):
...
     previous_vendor_id: Optional[str] = Field(
-        description="If specified, the unique identifier for the vendor that was previously associated with this system."
+        default=None,
+        description="If specified, the unique identifier for the vendor that was previously associated with this system.",
     )
-    dataset_references: List[FidesKey] = Field(
-        default_factory=list,
+    dataset_references: Optional[List[FidesKey]] = Field(
+        default=None,
         description="Referenced Dataset fides keys used by the system.",
```
pattisdr: Can we make default=[]? This seems a change, might affect when we're going to save systems to the backend.

```sql
                                             Table "public.ctl_systems"
                Column                |           Type           | Collation | Nullable |          Default          
--------------------------------------+--------------------------+-----------+----------+---------------------------
 dataset_references                   | character varying[]      |           | not null | '{}'::character varying[]
```

src/fideslang/models.py
```diff
@@ -1234,12 +1267,12 @@ def deprecate_fields(cls, values: Dict) -> Dict:
                 )
         return values
 
-    @validator("privacy_declarations", each_item=True)
+    @field_validator("privacy_declarations", check_fields=True)
```
pattisdr: I'm not following the `each_item` -> `check_fields`, can you explain? This validation no longer appears to be getting called.

tests/fideslang/test_validation.py
```diff
@@ -755,13 +740,11 @@ def test_data_categories_on_nested_fields(self):
...
         with pytest.raises(ValidationError):
-            CollectionMeta(after=[FidesCollectionKey("test_key")])
+            CollectionMeta(after=["test_key"])
 
     def test_collection_key_has_too_many_components(self):
         with pytest.raises(ValidationError):
-            CollectionMeta(
-                after=[FidesCollectionKey("test_dataset.test_collection.test_field")]
-            )
+            CollectionMeta(after=["test_dataset.test_collection.test_field"])
```
pattisdr, line 728: I'm assuming FidesCollectionKey could be removed here because this is already defined on the after field and would be used automatically.

src/fideslang/validation.py
```diff
@@ -3,50 +3,54 @@
...
+    return value
 
-        if not cls.regex.match(value):
-            raise FidesValidationError(
-                f"FidesKeys must only contain alphanumeric characters, '.', '_', '<', '>' or '-'. Value provided: {value}"
-            )
 
-        return value
+# Dataset.Collection name where both dataset and collection names are valid FidesKeys
+FidesCollectionKey = Annotated[str, PlainValidator(validate_collection_key_parts)]
```
pattisdr: I don't know if we had this behavior before, and lost in the refactor, but you're not prevented from passing in a non-string here, say an integer, which will fail when we go to run`value.split`

tests/conftest.py
pattisdr: ah ok, feels like something that will cause potential issues with existing yaml files etc.

tests/fideslang/test_validation.py
```diff
@@ -486,8 +470,8 @@ def test_invalid_country_identifier(country_code: str):
 def test_valid_country_identifier(country_code: str):
     """Validates usage of alpha-3 codes per ISO 3166"""
     System(
-        organization_fides_key=1,
-        registryId=1,
+        organization_fides_key="1",
+        registryId="1",
```
pattisdr: Casing of this is incorrect so is this field just ignored?

.github/workflows/pr_checks.yml
pattisdr: Understood, how do we generally communicate things like this to customers?

src/fideslang/validation.py
pattisdr: Thanks for all these helpful review comments

src/fideslang/models.py
```diff
@@ -1084,48 +1096,60 @@ class System(FidesModel):
     """
 
     registry_id: Optional[int] = Field(
```
pattisdr, line 1057: It looks like in FIdeslang the type _is_ an integer, but the tests updated to` registry_id="1"`

The `ctl_systems` table has it as a string though, so should this be updated to a string? Come to think of it I think I've run into 500 errors before around this - the discrepancy between fideslang and the database here.

```
                                            Table "public.ctl_systems"
                Column                |           Type           | Collation | Nullable |          Default          
--------------------------------------+--------------------------+-----------+----------+---------------------------
registry_id                          | character varying        |           |          | 
```

src/fideslang/validation.py
```diff
@@ -3,50 +3,54 @@
...
-class FidesVersion(Version):
-    """Validate strings as proper semantic versions."""
+def validate_fides_key(value: str) -> str:
+    """Throws ValueError if val is not a valid FidesKey"""
 
-    @classmethod
-    def __get_validators__(cls) -> Generator:
-        yield cls.validate
+    regex: Pattern[str] = re.compile(FIDES_KEY_PATTERN)
+    if not regex.match(value):
```
pattisdr, line 25: This throws confusing Type errors if the fides key is not a string, should we add more validation to require FidesKeys to be a string upfront?

tests/fideslang/test_models.py
pattisdr: I think we need to put this back and fix the model-level validation.

It looks like `privacy_declarations_reference_data_flows` is not getting called now.  This first test, the privacy declaration seems to reference a system but there are no system egresses?

pattisdr (commented):
Huge effort here Thomas.   Main thing I'd make sure these validators that are working on a list of items still work, in places they're not getting called, but only small things were noted here, quick to turn around.

Have you just pinned this latest fideslang commit in Fides and experimented with the level effort required here? I wonder if trying to integrate will cause more changes needed in Fideslang so it might be nice to verify before we do the big 3.0 release.

.github/workflows/pr_checks.yml
ThomasLaPiana: Other than changelog, it'll automatically fail to pip install if they have an older version.

Afaik most people are using Docker containers, this would be most disruptive to customers using the CLI

src/fideslang/validation.py
ThomasLaPiana: I copied you! haha yours are always so helpful

tests/conftest.py
ThomasLaPiana: potentially for sure...I'm going to test this branch out in Fides and see what kind of damage it does 😬

ThomasLaPiana:
> There's a whole new lingo to learn!

no kidding! Soooo many major changes here, hopefully the performance improvements and new ergonomics are worth it!

tests/conftest.py
ThomasLaPiana: it did, so I had to update everything in `Fides`....will need to think about how to handle this if we don't want to break things...

although coming from the db they should already be correct

src/fideslang/models.py
ThomasLaPiana: uh oh...hmm ok I'll need to do more testing over in `fides` for this

src/fideslang/validation.py
ThomasLaPiana: In which situations? We might need to pass in as `str` specifically

src/fideslang/models.py
ThomasLaPiana: this is kind of a messy/bad abstraction. Names can sometimes be optional, and sometimes are not (they're treated as keys) so in my opinion this was a premature optimization/abstraction on my part :)

src/fideslang/models.py
ThomasLaPiana: I don't have a good answer here tbh, I can't remember why/when we started adopting the new paradigm

src/fideslang/models.py
ThomasLaPiana: Seems like leftover tech-debt, I'm not sure either

src/fideslang/models.py
ThomasLaPiana: good point, done!

src/fideslang/models.py
ThomasLaPiana: This was a bad change on my part. Fixed in a more transparent validation method :)

src/fideslang/validation.py
ThomasLaPiana: I don't think this is an issue. Any failure inside of the validator should throw the same error, and the `str` type annotation should catch it I think anyway? Without doing `strict=False` Pydantic should throw an error instead of coercing, which is the opposite of how it worked before

tests/fideslang/test_validation.py
ThomasLaPiana: lol it would appear so....

NevilleS:
Migrated this one to Ethyca's downstream fork: https://github.com/ethyca/fideslang/pull/9. Once we finish this update for the Python bindings it'll be useful to merge upstream 👍

PR #162, opened 2023-09-08
Relax System Legal Basis for Transfers

pattisdr:
Closes N/A

### Description Of Changes

We keep encountering new values for system legal basis for transfers. For now, remove the enum and allow this value to be any string.  We'll rely on some frontend transformations to try to keep this in line, but this will avoid changes across multiple backend repos in the future.

### Code Changes

* [ ] legal basis for transfers changed to a list of strings

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Documentation Updated
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -229,7 +229,7 @@ class LegalBasisForTransfersEnum(str, Enum):
     The model for describing the legal basis under which data is transferred
 
     We currently do _not_ enforce this enum on the `legal_basis_for_transfers`
-    field, because the set of allowable values seems to be changing frequently 
+    field, because the set of allowable values seems to be changing frequently
     and without clear notice in upstream, public data sources.
     """
 
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -225,7 +225,13 @@ class LegalBasisForProfilingEnum(str, Enum):
 
 
 class LegalBasisForTransfersEnum(str, Enum):
-    """The model for describing the legal basis under which data is transferred"""
+    """
+    The model for describing the legal basis under which data is transferred
+
+    We currently do _not_ enforce this enum on the `legal_basis_for_transfers`
+    field, because the set of allowable values seems to be changing frequently 
+    and without clear notice in upstream, public data sources.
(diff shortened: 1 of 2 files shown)


src/fideslang/models.py
```diff
@@ -230,6 +230,7 @@ class LegalBasisForTransfersEnum(str, Enum):
     ADEQUACY_DECISION = "Adequacy Decision"
     SCCS = "SCCs"
     BCRS = "BCRs"
+    SUPPLEMENTARY_MEASURES = "Supplementary Measures"
```
pattisdr, line 239: Leaving this enum here, casing added to match Adequacy Decision - looks like FE does a transform

pattisdr:
Branched off of main here- assume we may need to follow up by cherry-picking onto 1.4.5 branch and release as 1.4.6 in the interim if needed quickly.

src/fideslang/models.py
```diff
@@ -225,11 +225,18 @@ class LegalBasisForProfilingEnum(str, Enum):
 
 
 class LegalBasisForTransfersEnum(str, Enum):
-    """The model for describing the legal basis under which data is transferred"""
+    """
+    The model for describing the legal basis under which data is transferred
+
+    We currently do _not_ enforce this enum on the `legal_basis_for_transfers`
```
adamsachs, line 231: @pattisdr what do you think? maybe this doesn't belong here, but i figured it would make it easier if/when we need to revisit this.

adamsachs (approved):
looks good, thank you!

i offered an additional note to a docstring, but would like your thoughts, i'm not sure about it 😄

src/fideslang/models.py
pattisdr: yep, great 👍  I didn't want to delete entirely yet

adamsachs: yes i agree with that! leaves things more traceable, since there's now going to be some discrepancies between our UI and the BE model.

and it's aspirational, i hope thing stabilize to a point where we _can_ leverage the enum here :)

PR #167, opened 2023-09-13
Base TCF Mapping

pattisdr:
### Description Of Changes

Adds mappings from GVL to new fideslang Purpose and Feature objects as well as utility functions from converting fideslang data uses to Purposes and vice versa. 

### Code Changes

* [ ] _list your code changes here_

### Steps to Confirm

### Pre-Merge Checklist

* [ ] Documentation Updated
* [ ] Update `CHANGELOG.md`


adamsachs (approved):
this looks great! love to see this finally getting into `main` here, it's been a long time coming :)

this _may_ be deserving of a `2.1.0` release? i don't think we're very formal with semantic versioning or anything, but adding in some GVL support seems like a bit of a milestone! 🤷

src/fideslang/gvl/gvl_feature_mapping.json
```diff
@@ -0,0 +1,32 @@
...
+            "description": "Data from offline data sources can be combined with your online activity in support of one or more purposes"
+        },
+        "2": {
+            "id": 2,
+            "name": "Link different devices",
+            "description": "Different devices can be determined as belonging to you or your household in support of one or more of purposes."
+        },
+        "3": {
+            "id": 3,
+            "name": "Receive and use automatically-sent device characteristics for identification",
```
pattisdr: I added these features from data I got from airtable, and double-checking here, I realize these names are from gvl v2. The purposes seem to be from gvl v3.  I assume I need to make sure everything here matches v3 - does that sound correct to you @adamsachs? 

Are we being consistent between v2 and v3 within airtable?

adamsachs: oh that's a great callout. i don't think we're being as consistent about this across our whole stack as we need to be.

i think we're _mainly_ using `v3` data within airtable and should aim to use `v3` until/unless it's causing us any problems. you're right that in the data use table in airtable we're currently using the v2 name here - i think that's mainly an artifact of me not keeping the data generation script well maintained.

i think we should update here to use the v3 name. i assume then we should update the data in airtable - both for general consistency and more concretely because it's the feature name (and not ID) that's being fed thru the dictionary and saved on our `System`s in fides...which you then use this mapping to reverse translate into an ID for the overlay...right?

this will all be a lot easier to keep consistent once we merge this PR, release `fideslang`, and then i switch the data generation script to leverage it for its mapping 😄

pattisdr: Correct yes, feature names need to match since they're saved on the system!

I'll update to make sure everything matches v3 here 👍

src/fideslang/gvl/gvl_data_use_mapping.json
```diff
@@ -73,7 +73,7 @@
             "id": 8,
             "name": "Measure content performance",
             "data_uses": ["analytics.reporting.content_performance"],
-            "description": "Information regarding which content is presented to you and how you interact with it can be used to determine whether the (non-advertising) content e.g. reached its intended audience and matched your interests. For instance, whether you read an article, watch a video, listen to a podcast or look at a product description, how long you spent on this service and the web pages you visit etc. This is very helpful to understand the relevance of (non-advertising) content that is shown to you. \n",
```
pattisdr: I removed some \n characters that are in the gvl mapping - I assume we don't want these here, but can revert if we want to match verbatim.

adamsachs (approved):
most recent changes look good, thanks for the diligence around the feature mapping 👍

src/fideslang/gvl/gvl_data_use_mapping.json
adamsachs: that seems fine to me, i'd just tried to copy verbatim from the GVL JSON source as a starting point here, but this isn't something we should need to be updating much, so removing `\n` seems like a fine additional step for now 👍

src/fideslang/gvl/gvl_feature_mapping.json
adamsachs: that sounds good, there's a slight chance we may need to revisit this if, for some reason, we decide we want to adhere to v2 for now (just because v3 still seems in the works, in some ways?)

but i don't want to hold this up, it's hard to get a definitive decision on all of that, let's shoot for v3 for now and adjust if/when we need to (hopefully we won't!)

PR #177, opened 2023-10-09
add support for PrivacyDeclaration.flexible_legal_basis_for_processin…

### LINKED ISSUE #176: add `flexible_legal_basis_for_processing` field

There's been discussion about anticipating the need to mark a given `PrivacyDeclaration` as having a _flexible_ legal basis, i.e. one that can be edited per-instance and is not purely determined by upstream sources.

**NOTE**: we'll also need a corresponding update (issue) in `fides` to update the DB model (and perform a migration) to get this change fully integrated!

add a new field, `flexible_legal_basis_for_processing`, to the `PrivacyDeclaration` model. default should be `False`.

hold off on making this change in fideslang (and fides) until we've solidified the requirement, if more solidification is needed!

### Additional context

see some internal comms for context: 
- https://ethyca.atlassian.net/wiki/spaces/EN/pages/2806415398?focusedCommentId=2806579295
- https://ethyca.slack.com/archives/C05E74TR0BB/p1696873548747899?thread_ts=1696870072.062149&cid=C05E74TR0BB

adamsachs:
Closes https://github.com/ethyca/fideslang/issues/176

### Description Of Changes

add the new field!

### Code Changes

* [x] added the new field!

### Steps to Confirm

n/a

### Pre-Merge Checklist

* [ ] Documentation Updated
* [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -992,7 +992,7 @@ class PrivacyDeclaration(BaseModel):
         default_factory=list, description="The features of processing personal data."
     )
     flexible_legal_basis_for_processing: bool = Field(
-        description="Whether the legal basis for processing is flexible (i.e. editable) for this declaration.",
+        description="Whether the legal basis for processing is 'flexible' (i.e. can be overridden in a privacy notice) for this declaration.",
         default=False,
     )
     legal_basis_for_processing: Optional[LegalBasisForProcessingEnum] = Field(
--- a/src/fideslang/models.py
+++ b/src/fideslang/models.py
@@ -991,6 +991,10 @@ class PrivacyDeclaration(BaseModel):
     features: List[str] = Field(
         default_factory=list, description="The features of processing personal data."
     )
+    flexible_legal_basis_for_processing: bool = Field(
+        description="Whether the legal basis for processing is flexible (i.e. editable) for this declaration.",
+        default=False,
+    )
     legal_basis_for_processing: Optional[LegalBasisForProcessingEnum] = Field(
         description="The legal basis under which personal data is processed for this purpose."
     )
--- a/tests/fideslang/test_models.py
+++ b/tests/fideslang/test_models.py
@@ -419,7 +419,8 @@ class TestSystem:
             uses_cookies=True,
             cookie_refresh=True,
             uses_non_cookie_access=True,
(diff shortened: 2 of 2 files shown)


src/fideslang/models.py
```diff
@@ -991,6 +991,10 @@ class PrivacyDeclaration(BaseModel):
     features: List[str] = Field(
         default_factory=list, description="The features of processing personal data."
     )
+    flexible_legal_basis_for_processing: bool = Field(
+        description="Whether the legal basis for processing is flexible (i.e. editable) for this declaration.",
```
NevilleS:
```suggestion
        description="Whether the legal basis for processing is 'flexible' (i.e. can be overridden in a privacy notice) for this declaration.",
```
Slightly better phrasing, but that's about the best we can do here! It's always going to be a mouthful.

src/fideslang/models.py
```diff
@@ -991,6 +991,10 @@ class PrivacyDeclaration(BaseModel):
     features: List[str] = Field(
         default_factory=list, description="The features of processing personal data."
     )
+    flexible_legal_basis_for_processing: bool = Field(
```
NevilleS, line 994: I know it might seem nitpicky, but can this be an `Optional[bool]`, is that a thing? Having it possible to be completely `None` feels better than a default of `false` to me.

NevilleS (approved):
Nitpicks inline, but this works! 👍

src/fideslang/models.py
adamsachs: it's overriden in the notice? sorta curious what that means, but i will go with your suggestion here 👍

src/fideslang/models.py
adamsachs: hmmm. i see what you mean, but our standard/convention on this model here is already to default to `False` and i _do_ think that we want to basically assume `False` as our default behavior here, at least at the application level. 

so i think we'll want/need a default somewhere in our stack anyway, the flexibility to have an empty value could be nice, but i'm worried that it'll add more edge case/complexity than it's worth at this point. unless there's a really strong reason we want to keep that empty state as a possibility here.

so...i'm gonna forego this suggestion, if that's ok 😄

PR #182, opened 2023-11-06
support `previous_vendor_id` attribute on System model

adamsachs:
part of https://ethyca.atlassian.net/browse/PROD-1296

### Description Of Changes

Adds a new, optional `previous_vendor_id` attribute/field to the `System` model. This field will help us indicate whether a given vendor record (or a system based on that vendor) used to be represented by a different vendor record, e.g. if a google AC vendor transitions to a GVL vendor, as is expected to happen somewhat regularly.

### Code Changes

* [x] add `previous_vendor_id` attribute to `System` model
* [x] some test cleanup, add a TODO comment since i realized some of our model tests aren't really effective...

### Steps to Confirm

* [x] will test out by publishing an alpha tag and using this in Compass webservice code to ensure it can support this new column on Compass data. UPDATE: tested with 2.2.1a0 alpha package, works well 👍 

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`

### DIFF
--- a/tests/fideslang/test_models.py
+++ b/tests/fideslang/test_models.py
@@ -92,6 +92,10 @@ class TestPrivacyDeclaration:
 
 
 class TestSystem:
+    # TODO: these tests are not effectively evaluating whether the provided constructor args
+    # are actually supported, because our `System` model does not prohibit "extra" fields.
+    # We need to update these tests to assert that the provided args are actually being set
+    # as attributes on the System instance that's instantiated.
     def test_system_valid(self) -> None:
         assert System(
             description="Test Policy",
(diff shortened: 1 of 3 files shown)


tests/fideslang/test_models.py
```diff
@@ -419,7 +424,7 @@ def test_expanded_system(self):
             cookie_refresh=True,
             uses_non_cookie_access=True,
             legitimate_interest_disclosure_url="http://www.example.com/legitimate_interest_disclosure",
-            flexible_legal_basis_for_processing=True,
```
adamsachs, line 422: this is unrelated to the change in this PR, but saw this test mistake while i was updating the test. i moved this up to the `PrivacyDeclaration` where it should be, this was just a mindless mistake to have it on the `System` in the first place.

more importantly, the test has been passing even with this error in here, which means that the test is ineffective! i put in a TODO comment for now, i'll see if i can circle back to improving the test...

tests/fideslang/test_models.py
```diff
@@ -390,13 +394,14 @@ def test_expanded_system(self):
...
                     shared_categories=[],
+                    flexible_legal_basis_for_processing=True,
                     cookies=[
                         {"name": "ANON_ID", "path": "/", "domain": "tribalfusion.com"}
                     ],
                 )
             ],
             third_country_transfers=["ARM"],
-            vendor_id="1",
+            vendor_id="gvl.1",
```
adamsachs, line 404: this doesn't really matter but figured i'd update to make the value more realistic in our test case

pattisdr (approved):

PR #186, opened 2023-12-08
Fideslang 3.0

adamsachs:
Closes [PROD-1490](https://ethyca.atlassian.net/browse/PROD-1490)

### Description Of Changes

Updates to remove generally unused/deprecated concepts. This will be a significant breaking change, so it justifies a major upgrade, i.e. `3.0`.

### Code Changes

- [x] remove data qualifiers
- [x] remove deprecated fideslang fields (a lot of them are from the Compass updates - they are marked as deprecated)
- [x] remove the registry concept - optional

### Steps to Confirm

* [x] extensive integration testing with `fides` and `fidesplus`!
* [x] update docs and references for CLI and general taxonomy 
* [ ] ...

### Pre-Merge Checklist

* [ ] Documentation Updated
* [x] Update `CHANGELOG.md`

[PROD-1490]: https://ethyca.atlassian.net/browse/PROD-1490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/fideslang/models.py
```diff
@@ -687,53 +577,20 @@ class Dataset(FidesModel, FidesopsMetaBackwardsCompat):
...
-    data_qualifier: Optional[FidesKey] = Field(
-        description="Deprecated. Array of Data Qualifier resources identified by `fides_key`, that apply to all collections in the Dataset.",
-    )
     fides_meta: Optional[DatasetMetadata] = Field(
         description=DatasetMetadata.__doc__, default=None
     )
-    joint_controller: Optional[ContactDetails] = Field(
-        description="Deprecated. " + ContactDetails.__doc__,
-    )
-    retention: Optional[str] = Field(
```
adamsachs, line 699: removed from top-level `Dataset` but kept it on `DatasetCollection` and `DatasetField` since it hadn't been marked for deprecation there...

src/fideslang/models.py
```diff
@@ -1150,19 +942,10 @@ class System(FidesModel):
...
-        description="Deprecated. " + ContactDetails.__doc__,
-    )
-    third_country_transfers: Optional[List[str]] = Field(
-        description="Deprecated. An optional array to identify any third countries where data is transited to. For consistency purposes, these fields are required to follow the Alpha-3 code set in ISO 3166-1.",
-    )
     administrating_department: Optional[str] = Field(
         default="Not defined",
         description="An optional value to identify the owning department or group of the system within your organization",
     )
-    data_protection_impact_assessment: Optional[DataProtectionImpactAssessment] = Field(
```
adamsachs, line 1163: still kept `System.requires_data_protection_assessments` since it hadn't been marked for deprecation and it seems like it _maybe_ could still be useful?

adamsachs (commented):
@pattisdr would you mind taking an initial look at this and associated fides PR https://github.com/ethyca/fides/pull/4502?

i still will need to do some more testing, and i'll need to work in a fidesplus PR and make any adjustments as needed there, but i'd like to get your initial thoughts to make sure there's nothing sticking out as missing!

pattisdr:
Starting review...

src/fideslang/models.py
pattisdr: I think we should remove them too?   We should run by Neville but if memory serves, he was advocating that "datasets" should have legal concepts removed from them and they should become purely a "data engineering" concept.

src/fideslang/models.py
pattisdr: That makes sense to me.  I actually _added_ this field as as part of deprecating `data_protection_impact_assessment` https://github.com/ethyca/fideslang/pull/146.  I migrated over this value from System.data_protection_impact_assessment - Rachel didn't know if it would be needed or not, so we erred on preserving that data.

> System.requires_data_protection_assessments
> New field: Migrate value from System.data_protection_impact_assessment.is_required or set default value, then make non-nullable

pattisdr (commented):
Nice Adam 👍

I'm so happy you're deprecating registries too. It's a cause of 500 errors when patching Systems via the API directly because there's actually a mismatch between the database type and fideslang type - Registries was never used as intended and it just causes problems from time to time!

adamsachs:
marking as ready for review since i've now done a fair amount of testing with fides and fidesplus and generally things are looking good, feeling comfortable that this is a stable update that has what we need 👍 

one thing i need to follow up on - anything to check for generated docs/webpages/visuals? any guidance that you have @pattisdr is appreciated, but i can also dig a bit!

pattisdr (approved):
This is great 🏆

CHANGELOG.md
pcoatszsu, line 1: @adamsachs Hi, can you point me to the rationale for removing the 4th classification group from the ontology? I still see 4 referenced at certain places in the documentation. Thanks in advance for your attention!

